PR #2, opened 2022-07-01
Move stub package to its own folder

mahesh-hegde:
Just a series of git renames.


mahesh-hegde:
Update: I realized I forgot to modify the .github/workflows to run in jnigen/ directory instead; what is an idiomatic way to do it?

.github/workflows/test-package.yml
```diff
@@ -64,10 +64,12 @@ jobs:
         run: dart pub get
       - name: Run VM tests
         run: dart test --platform vm
+      - name: Install coverage
+        run: dart pub global activate coverage
       - name: Collect coverage
-        run: ./tool/coverage.sh
```
dcharkes, line 62: Then we should also delete that fine, not move it.

mahesh-hegde: Haha newbie mistake. Fixed it.

mahesh-hegde:
Is it sufficient to `grep -r` for jni_gen when renaming the package? I think it takes some attempts to get it right.

.github/workflows/test-package.yml
```diff
@@ -43,6 +46,9 @@ jobs:
   test:
     needs: analyze
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: ./jnigen
```
dcharkes, line 51: here's another one for jni_gen

mahesh-hegde:
Sorry too many little mistakes.

.github/workflows/test-package.yml
```diff
@@ -21,6 +21,9 @@ jobs:
   # against Dart stable.
   analyze:
```
dcharkes, line 22: We will have two sub-packages jni_gen and jni.
We should analyze and test both.

So we should either (a) call this analyze_jni_gen or (b) move the `working-directory` to each run step, so we can have a different working directory for analyzing jni.

Though we can also address this with the PR adding `jni`.

dcharkes:
> Sorry too many little mistakes.

Don't worry about it!

.github/workflows/test-package.yml
mahesh-hegde: working-directory is per-job it seems. I think I will add an analyze_jni in the commit adding jni. jni_gen is the main part anyway.

mahesh-hegde: Nvm. Renamed it for uniformity.

dcharkes:
> The workflow is not valid. .github/workflows/test-package.yml (Line: 47, Col: 12): Job 'test_jni_gen' depends on unknown job 'analyze'.

mahesh-hegde:
Statically typed YAML when? :P

dcharkes:
Well, you'll be a Dart package CI expert in no-time this way 😄

dcharkes (approved):
LGTM!

devoncarew:
Hi - as part of this change, we'll also want to delete the CONTRIBUTING.md file, move the authors file to the top level of the repo, and copy the license file to the top level (in addition to creating a top-level readme file).

Here's an example of a top-level readme file for a monorepo: https://github.com/google/protobuf.dart. It has a brief description of the repo and uses a markdown table to direct people to the individual contained packages.

dcharkes:
Oops, looks like the contributing.md got revived by the merge.

@mahesh-hegde would you please make a follow up PR to address these comments.

mahesh-hegde:
While we are at it, should I also link the proposal from README?

PR #9, opened 2022-07-07
Add jni

mahesh-hegde:
* Adds JNI to FFI plugin, along with android support

* There are several generated files which can be hard to review.
* To verify reproducibility

  + run tool/gen_aux_methods.dart in jni/ folder

  + run `dart run ffigen --config ffigen.yaml` to generate bindings to jni.h ; The header file is copied from Android NDK plus some annotations for param names in function pointer fields. The `ffigen` currently links to my fork of ffigen with a [jni-specific patch](https://github.com/mahesh-hegde/ffigen_patch_jni), on which changes from upstream `ffigen` should be easy to verify. I think that's better than committing all vendored files in the same commit.

Marking it as draft to get the CI right, eventually.


mahesh-hegde:
Closing this P

Issue #15, opened 2022-07-14
Windows support


mahesh-hegde:
This is a tracking issue for windows support.

`dart test` runs on windows after fixing error 126 with DELAYLOAD option. The example application also works.

However

* CI cannot run windows due to an issue with git dependency.
* `flutter test` failed, I will try again to reproduce it.

Creating a tracking issue and assigning to myself.

mahesh-hegde:
This is not as simple as just setting DELAYLOAD linker flag. I recently tried to run the same commit (`HEAD` of `add_jni` branch) on a different windows setup. It failed with DLL loading error again.

I doubt there's even a single point of failure. It can be CMake, it can be JDK, it can be particular windows SDK, VS installation etc..

I have to search for other libraries using JNI on windows.

mahesh-hegde:
#59 covers some of it.

The pain point is, at least the JDK I have (Microsoft build of OpenJDK 11) doesn't place jvm.dll in `%JAVA_HOME%\bin\`  but in `%JAVA_HOME%\bin\server`.

So dartjni.dll fails to load, not because the library is faulty, but it can't load jvm.

The solution is to add `%JAVA_HOME%\bin\server` to PATH, but that has to be done manually. We cant set env var from dart code.

So that means:

* If you want to run JNI tests on windows dev machine, just do this.
* `$env:Path += ";${env:JAVA_HOME}\bin\server"`. (If JAVA_HOME not set, find where's your `java` executable and set that in control panel).
* Now it should be possible to run tests and examples on standalone and flutter targets.

That's one step more than linux. Not sure if I can tell windows is _supported_.

mahesh-hegde:
I am developing on a Linux machine as well as windows machine for quite a while now. There are also CI tests for windows. This can be closed.

The only quirk is env var.

PR #19, opened 2022-08-10
Initial code generator support draft

mahesh-hegde:
This PR adds initial support for code generator.

There are 2 parts;

    * ApiSummarizer: Java component that generates JSON representation of Java APIs.
    * jni_gen: takes JSON and generates code based on it.

test package is in jni_gen_test, which tests generating code for a very simple java class.

Locally I am able to generate code for some PDFBox classes and some packages from android libraries as well. Still contemplating how to add these to CI. PDFBox requires 5-6 jar dependencies to be downloaded from apache and bouncy castle sites. Looking for a way to automate it with maven or something. Similarly `android-stubs-src.jar` from SDK-28 is the one working with Android libraries.

TODO: (Before merging this PR)
* How to add a LICENSE to `ApiSummarizer`? It depends on some JDK APIs, should it be GPL?


dcharkes:
> * How to add a LICENSE to `ApiSummarizer`? It depends on some JDK APIs, should it be GPL?

I believe this is the case. See https://www.gnu.org/licenses/gpl-faq.en.html#IfLibraryIsGPL. This is why LGPL was invented, it allows linking/importing a LGPL library in non-LGPL code. If the APIs are GPL, we are out of luck and have to use GPL. cc @mit-mit.
(This is hairy, if we load a `.jar` in a Java VM that we spun up with `dart:ffi` & JNI, is that considered linking as well? In that case the GPL would spill to this whole package. Maybe then it is a good thing that the API summarizer is Java and standalone. Then the Dart program that takes the summary doesn't have to link to that, and that can then be released under the Dart license.)

@devoncarew we're adding Java code to a Dart package. Do we have a practise for what Java package name to use? Something like `dev.pub.jni_gen.<...>`? Or `com.github.dart_lang.jni_gen.<...>` (In the Flutter plugins we have `com.example.<package-name>` which gives no guidance to the user.)

jni_gen_test/.gitignore
```diff
@@ -0,0 +1,16 @@
+# Files and directories created by pub.
```
dcharkes, line 1: Can `jni_gen_test` be `jni_gen/test` instead?

It should be the unit tests and integration tests for `package:jni_gen`.

jni_gen_test/run_tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
dcharkes, line 1: Prefer using a Dart script. E.g. a `setup.dart` with `Process.run`s.

dcharkes, line 1: You could invoke the setup script from the `setUp` in the test, so it is automatically run.
Be sure to print that is running as well for if the setup fails.

jni_gen_test/test/jni_gen_test_test.dart
```diff
@@ -0,0 +1,40 @@
...
+
+  test('static final fields', () {
+    expect(Example.ON, equals(1));
+    expect(Example.OFF, equals(0));
+  });
+
+  test('static & instance fields', () {
+    expect(Example.num, equals(121));
+
+    final aux = Example.aux;
```
dcharkes, line 21: Generate a comment on the `aux` getter it creates a reference from Dart to Java that must be `delete`d later.

jni_gen_test/lib/dev/dart/jni_gen_test.dart
```diff
@@ -0,0 +1,144 @@
...
+class Example extends jni.JlObject {
+  Example.fromRef(ffi.Pointer<ffi.Void> ref) : super.fromRef(ref);
+
+  /// from: static public final int ON
+  static const ON = 1;
+
+  /// from: static public final int OFF
+  static const OFF = 0;
+
+  /// from: static public dev.dart.jni_gen_test.Example.Aux aux
```
dcharkes, line 25: This comment is not on `static Example_Aux get aux` but on a private member.

jni_gen_test/test/jni_gen_test_test.dart
```diff
@@ -0,0 +1,40 @@
...
+    expect(aux.value, equals(true));
+    aux.delete();
+  });
+
+  test('static methods', () {
+    expect(Example.addInts(10, 15), equals(25));
+  });
+
+  test('instance methods', () {
+    final ex = Example();
```
dcharkes, line 31: Idem here. The dartdoc for `Example()` should document it must be `delete`d.

jni_gen_test/tool/gen.dart
```diff
@@ -0,0 +1,24 @@
...
+    await runTask(JniGenTask(
+        summarySource: SummarizerCommand(
+          sourcePaths: [Uri.directory(javaPath)],
+          classPaths: [Uri.directory(javaPath)],
+          classes: ['dev.dart.jni_gen_test.Example'],
+        ),
+        outputWriter: FilesWriter(
+            cWrapperDir: Uri.directory('src/'),
+            dartWrappersRoot: Uri.directory('lib/'),
+            libraryName: 'jni_gen_test')));
```
dcharkes, line 18: nit: Add comma's between the closing brackets so that the formatting becomes all 2 spaces indented.

It makes the code easier to read.

jni_gen_test/tool/gen.dart
```diff
@@ -0,0 +1,24 @@
...
+  } on Error catch (e) {
+    stderr.writeln(e);
```
dcharkes, line 20: This doesn't print the stacktrace, correct? Just the toString() of the exception or error.

Maybe don't catch them at all, then they will just be uncaught and reported to the user with all available information.

jni_gen_test/test/jni_gen_test_test.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'package:jni_gen_test/dev/dart/jni_gen_test.dart';
+import 'package:test/test.dart';
+
+void main() {
+  if (!Platform.isAndroid) {
+    Jni.spawn(helperDir: 'build/jni_libs', classPath: ['java/']);
+  }
+
+  test('static final fields', () {
+    expect(Example.ON, equals(1));
```
dcharkes, line 14: This is super cool to see! 🔥

jni_gen_test/java/dev/dart/jni_gen_test/Example.java
```diff
@@ -0,0 +1,53 @@
...
+		num = 121;
+	}
+
+	// static methods
+	public static Aux getAux() {
+		return aux;
+	}
+
+	public static int addInts(int a, int b) {
+		return a+b;
```
dcharkes, line 23: Is this default Java formatting? No space between the + and its operands? Please run a formatter.

(Does Java really use tabs by default to format. 😭 Can we change it to spaces? 💪 )

jni_gen_test/tool/gen.dart
```diff
@@ -0,0 +1,24 @@
+import 'dart:io';
+import 'package:jni_gen/jni_gen.dart';
+
+void main(List<String> arguments) async {
```
dcharkes, line 4: This should be a command-line interface in `jni_gen/bin/jni_gen.dart` so that we can do `dart run jni_gen --config <some-config-file>.yaml`.

For a test that uses the Dart API instead of a config file: We should in a test regenerate the file and check that it is equal to what we had before. (Give the `_test.dart` file a parameter to keep the temporary files, so we can easily inspect the API summary as well.)

jni_gen/bin/build_summarizer.dart
```diff
@@ -0,0 +1,53 @@
+// this script takes a vendored ApiSummarizer and builds with maven.
```
dcharkes, line 1: should this be `bin/setup.dart` instead?

dcharkes, line 1: this -> This

jni_gen/cmake/CMakeLists.txt.tmpl
```diff
@@ -0,0 +1,30 @@
+# The Flutter tooling requires that developers have CMake 3.10 or later
```
dcharkes, line 1: Remove this. We're not using this in Flutter apps.

dcharkes (commented):
First round of comments. I'll take a look at the Java code later.

dcharkes:
Super cool stuff! 🔥 🎸 🚀

mahesh-hegde:
The doclet API has same license as rest of the java standard library, for example compare [java.io.FileOutputStream](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/io/FileOutputStream.java) and [jdk.javadoc.doclet.Doclet](https://github.com/openjdk/jdk/blob/master/src/jdk.javadoc/share/classes/jdk/javadoc/doclet/Doclet.java).

On surface, we are using an API with classpath exception. However, the API calls the javadoc tool itself, I am not sure if such use is covered under classpath exception or not.

It's possible I am being overly cautious. There are some uses similar to ours covered under permissive licenses.

[Selenium](https://sourcegraph.com/github.com/SeleniumHQ/selenium/-/blob/java/src/dev/selenium/tools/javadoc/JavadocJarMaker.java)
[Scala SBT](https://sourcegraph.com/github.com/SeleniumHQ/selenium/-/blob/java/src/dev/selenium/tools/javadoc/JavadocJarMaker.java)

jni_gen/bin/build_summarizer.dart
mahesh-hegde: Is `setup.dart` a common convention? If so I will rename it.

jni_gen_test/java/dev/dart/jni_gen_test/Example.java
mahesh-hegde: I should have run google-java-format on it (As rest of the summarizer).

jni_gen/bin/build_summarizer.dart
dcharkes: More common than a specific name. The `bin/*.dart` files are not easily discoverable. `dart run jni_gen:setup` seems clearer in docs to me than `dart run jni_gen:build_summarizer`. The users of `package:jni_gen` do not have to know that we have two separate components on the inside (the summarizer and generator). They just want to run one setup step and be done.

jni_gen_test/java/dev/dart/jni_gen_test/Example.java
dcharkes: Maybe run the formatter on CI as well, that way we don't break it again later. 🥇

jni_gen_test/.gitignore
mahesh-hegde: jni_gen_test is a complete dart package with some tests after wrapper code is generated. Is it fine to nest a package like that in test/ folder?

jni_gen_test/run_tests.sh
mahesh-hegde: I am actually not using it anywhere but local machine :). On CI, each `dart run` is separate named step.

jni_gen_test/tool/gen.dart
mahesh-hegde: Is `golden file` testing viable at this point? Because generated code structure is still in flux.

jni_gen_test/.gitignore
dcharkes: Why do you need a package?

* Any dependencies only use in tests can go in dev dependencies
* The wrapper code can be generated in the `setUp` of the test.

Is there something I'm missing?

jni_gen_test/run_tests.sh
dcharkes: Well, I was running it as well, because I did not want to have to figure out in which order to do things and what steps to run! 😆 

It would be nice if we could just `cd jni_gen`, `dart test` and then everything needed for the setup of the tests is automatically run.

jni_gen/cmake/CMakeLists.txt.tmpl
mahesh-hegde: `FilesWriter` copies this along with generated C file, because generated library needs to be linked to JVM libraries etc.. depending on the platform. This applies to both flutter and standalone.

jni_gen_test/.gitignore
dcharkes: If you want full-package integration tests, you can take a look at how they are setup in:

https://github.com/dart-lang/ffigen/tree/master/example/c_json
https://github.com/dart-lang/ffigen/blob/master/test/large_integration_tests/large_test.dart#L62
https://github.com/dart-lang/ffigen/blob/master/test/large_integration_tests/_expected_cjson_bindings.dart

jni_gen/cmake/CMakeLists.txt.tmpl
dcharkes: I'm blind and didn't see the `.tmpl` 🙈

jni_gen_test/tool/gen.dart
dcharkes: Yeah I think it's fine. It also makes it easy to see how we change our generated code on subsequent PRs.

jni_gen_test/.gitignore
mahesh-hegde: > Why do you need a package?

The details are now slightly fuzzy in my head. But you can't generate code and use that in single test file. Test file that calls generated code should be loaded _after_setup is run.

Also code in `test/` is analyzed by `dart analyze` which may catch code depending on yet-to-be-generated sources.

I will see if there's another way to do this.

> large_test.dart

I think it only compares with expected bindings and does not run anything based on generated code. I wanted something end-to-end, that runs some test on generated code.

At this point jni_gen is at very early version and I think it's not viable to rely on exact comparison of generated bindings.

jni_gen_test/tool/gen.dart
mahesh-hegde: In `ffigen` I saw there was quite a bit of code dedicated for YAML parsing. Is there now something like `json_serializable` for YAML? I think YAML config would not be an immediate priority.

jni_gen_test/.gitignore
dcharkes: > But you can't generate code and use that in single test file.

Correct, but you can (1) commit the generated code and generate code and use it in tests, and then (2) run the generator and compare the output string with the committed generated code in a second test.

In general, you cannot have test referring to generated files and not committing the generated file, even if you do it in a setup step.

> At this point jni_gen is at very early version and I think it's not viable to rely on exact comparison of generated bindings.

I think it's fine to update generated files every PR. This is what we have done with the `libclang` bindings in `ffigen` as well.

jni_gen_test/tool/gen.dart
dcharkes: We can leave the config for later.

In that case we will not have a `bin/jni_gen.dart` for now.

We will only have a `lib/jin_gen.dart` that exposes a Dart API that is directly used by our tests.

jni_gen_test/java/dev/dart/jni_gen_test/Example.java
mahesh-hegde: It will require to either download the jar from google-java-format release or use a third party GH action (axel-op/googlejavaformat-action).

Which one is preferred?

dcharkes: The easiest one. (Which is the GitHub action I would think, because it downloads the jar for us.)

jni_gen_test/.gitignore
liamappelbe: Perhaps this is more of an example than a test? In that case you could put it in jni_gen/example? You can still run an example on github CI as if it was an integration test, it just won't be picked up by `dart test`

jni_gen/lib/src/util/find_package.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'package:package_config/package_config.dart';
+
+Future<Uri?> findPackage(String packageName) async {
+  final packageConfig = await findPackageConfig(Directory.current);
+  if (packageConfig == null) {
+    return null;
+  }
+
+  final packages = packageConfig.packages;
+  for (var candidate in packages) {
```
liamappelbe, line 12: Can you replace this loop with this function? https://pub.dev/documentation/package_config/latest/package_config.package_config_types/PackageConfig/operator_get.html

jni_gen/lib/src/util/name_utils.dart
```diff
@@ -0,0 +1,27 @@
...
+/// splits [str] into 2 from last occurence of [sep]
+List<String> cutFromLast(String str, String sep) {
+  final li = str.lastIndexOf(sep);
+  if (li == -1) {
+    return ['', str];
+  }
+  return [str.substring(0, li), str.substring(li + 1)];
+}
+
+String getSimpleName(String binaryName) => binaryName.split('.').last;
```
liamappelbe, line 14: Duplicate of the `getLastName` function below?

jni_gen/lib/src/util/name_utils.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:jni_gen/src/elements/elements.dart';
+
+String getPackageName(String binaryName) => cutFromLast(binaryName, '.')[0];
+
+/// splits [str] into 2 from last occurence of [sep]
+List<String> cutFromLast(String str, String sep) {
```
liamappelbe, line 6: It's a shame that Dart doesn't have multi-return. Returning a List is pretty gross, but the only other option is writing a one-time-use class for the return type. A class would be cleaner, but I'll leave it up to you to decide if it's worth the effort.

jni_gen/lib/src/util/rename_conflict.dart
```diff
@@ -0,0 +1,54 @@
...
+String kwRename(String name) {
+  if (_keywords.contains(name)) {
+    return '${name}_';
+  }
+  return name;
+}
+
+// TODO: Add keywords(C) - keyword(Java)
+// TODO: change C constructor naming from 'new' to something else?
+const Set<String> _keywords = {
```
liamappelbe, line 22: Are these Dart's keywords? There's quite a few missing. Did you add Dart's keywords but then exclude Java keywords or something?

jni_gen/third_party/ApiSummarizer/todo.md
```diff
@@ -0,0 +1,10 @@
+* ASM
```
liamappelbe, line 1: Move all these to github issues?

jni_gen/lib/src/util/rename_conflict.dart
mahesh-hegde: Yes I was lazy and excluded Java's keywords because they cannot be in Java's identifier names.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/elements/TypeDecl.java
```diff
@@ -0,0 +1,12 @@
+package com.github.hegde.mahesh.apisummarizer.elements;
+
+public interface TypeDecl {
+  // can be EnumType, ClassType, InterfaceType, or ArrayType
```
dcharkes, line 4: Punctuation. Start with a capital, and with a dot.

(Or remove the comment, the DeclKind is an enum, so it's self documenting what kinds it can be.)

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/elements/TypeDecl.java
```diff
@@ -0,0 +1,12 @@
...
+  DeclKind getKind();
+
+  String getQualifiedName();
+
+  String getSimpleName();
+  // String getPackageName();
```
dcharkes, line 10: Don't commit commented out code.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/elements/JavaDocComment.java
```diff
@@ -0,0 +1,14 @@
+package com.github.hegde.mahesh.apisummarizer.elements;
+
+public class JavaDocComment {
+  public String comment;
+
+  // TODO: Escape HTML tags, and optionally convert the familiar ones to markdown.
+
+  // TODO: Build a detailed tree representation of JavaDocComment
```
dcharkes, line 8: File a bug on this repo, and link to the bug. (Idem other TODOs. Ctrl+f TODO and see if all have a link to a GitHub issue. 😄)

Reasoning for TODOs on GitHub instead of in code: it is easier to discuss alternative solutions and if we want to prioritize it on a GitHub issue. Also a GitHub issue has more space for elaborating (instead of leaving a 10 line TODO in the code).

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/elements/TypeUsage.java
```diff
@@ -0,0 +1,76 @@
...
+  public enum Kind {
+    DECLARED,
+    TYPE_VARIABLE,
+    WILDCARD,
+    ARRAY,
+    INTERSECTION,
+    PRIMITIVE,
+  }
+
+  // Could've made made it just a type hierarchy
```
dcharkes, line 15: Punctuation

```
  // Could've made made it just a type hierarchy, but client code parsing JSON often needs
  // to know the type beforehand.
```

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/util/SkipMethodException.java
```diff
@@ -0,0 +1,7 @@
+package com.github.hegde.mahesh.apisummarizer.util;
+
+public class SkipMethodException extends RuntimeException {
```
dcharkes, line 3: Should this extend SkipException?

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/util/SkipTypeException.java
```diff
@@ -0,0 +1,7 @@
+package com.github.hegde.mahesh.apisummarizer.util;
+
+public class SkipTypeException extends RuntimeException {
```
dcharkes, line 3: Should this extend SkipException?

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/doclet/AnnotationVisitor.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  @Override
+  public Object visitType(TypeMirror typeMirror, Void unused) {
+    return builders.typeUsage(typeMirror);
+  }
+
+  @Override
+  public Object visitEnumConstant(VariableElement variableElement, Void unused) {
+    // TODO: Do this properly
```
dcharkes, line 78: This is not self-documenting. What is not proper about it? Please file a GitHub issue.

nit: punctuation.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/doclet/ElementBuilders.java
```diff
@@ -0,0 +1,215 @@
...
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import javax.lang.model.element.*;
+import javax.lang.model.type.*;
+
+public class ElementBuilders {
+  AstEnv env;
+
+  private static <T, R> List<R> collect(List<T> elements, Function<T, R> mapper) {
```
dcharkes, line 16: The method collect(List<T>, Function<T,R>) from the type ElementBuilders is never used locally.

Maybe run a Java analyzer over the codebase.
Can we also run a Java analyzer on the CI?

Here's what my IDE is reporting: https://gist.github.com/dcharkes/77f3ec4c99f4a4f63faed385a1817c60

dcharkes (commented):
Super nice to play around with the summarizer 🚀 

More comments 😄

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/Main.java
```diff
@@ -0,0 +1,192 @@
...
+          files.add(path);
+        }
+      }
+    }
+    return files;
+  }
+
+  public static CommandLine parseArgs(String[] args) {
+    var options = new Options();
+    Option asm =
```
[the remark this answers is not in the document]
mahesh-hegde: I toyed around a bit with ASM library (I had also written some ASM code while writing the prototype for this project), but it doesn't work yet. I will get to this option if & when ASM is implemented.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/elements/ClassDecl.java
```diff
@@ -0,0 +1,31 @@
...
+
+// Currently Interfaces are stored as ClassDecl
+// EnumDecl is a subclass which has names of enum constants
+
+/** Denotes class or interface declaration. */
+public class ClassDecl {
+  public DeclKind declKind;
+  public Set<String> modifiers;
+  public String qualifiedName;
+  public String simpleName;
```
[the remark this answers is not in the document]
mahesh-hegde: Qualified name is not binary name. Binary name will be unique, qualified name is not.

That said, I think I will do away with both qualName and internalName. BinaryName is enough.

ApiSummarizer's ElementBuilder parts are some of the highly boilerplate parts, so I just listed out all the properties I could see at once and wrote builders for them in flow. :P

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/util/SkipTypeException.java
mahesh-hegde: I think I can do away with SkipMethodException and SkipClassException, because realistically only handler will know to which extent to skip.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/Main.java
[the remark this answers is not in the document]
dcharkes: Then just fully remove the code from the PR for now.

(Be sure to save it somewhere on a branch or something.)

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/doclet/ElementBuilders.java
mahesh-hegde: Correct. At some point replaced it with `StreamUtils.mapped`.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/elements/ClassDecl.java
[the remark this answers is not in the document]
dcharkes: Now that we have a generator, we can prune the JSON summary to only contain the things we use (or will likely use in the future).

jni_gen/test/sample/.gitignore
```diff
@@ -0,0 +1,5 @@
+build/
```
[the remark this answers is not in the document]
mahesh-hegde: I just thought that's neater than littering the test folder with subdirectories like lib/ and src/. Now what's the best place to keep it?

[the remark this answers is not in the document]
dcharkes: If the example is useful for people as documentation, put it in a top-level example folder.

jni_gen/third_party/ApiSummarizer/src/main/java/com/github/hegde/mahesh/apisummarizer/Main.java
```diff
@@ -0,0 +1,192 @@
...
+        new Option("A", "use-asm", false, "Use ASM based summarizer and search class path");
+    Option sources = new Option("s", "sources", true, "paths to search for source files");
+    Option classes = new Option("c", "classes", true, "paths to search for compiled classes");
+    Option useModules = new Option("M", "use-modules", false, "use Java modules");
```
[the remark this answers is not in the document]
mahesh-hegde: From java 9 there's a feature called modules which changes package layout if used. I am just forwarding this option to javadoc tool. It seems it is required to generate javadoc for stdlib but didnt figure out how.

mahesh-hegde:
Hi, I think I implemented most of the suggestions. What do we do about the license?

dcharkes:
> Hi, I think I implemented most of the suggestions. What do we do about the license?

Java code GPL. And it has to stay in third_party.

The generator can be the Dart license. It is very handy that we have the summarizer as a separate program now. Because we don't link it from the Dart, we just invoke it with `Process.run` or some other script.

.github/workflows/test-package.yml
```diff
@@ -76,42 +82,23 @@ jobs:
           parallel: true
           path-to-lcov: ./jni_gen/coverage/lcov.info
 
-  test_jni_gen_test_plugin:
-    needs: test_jni_gen
-    runs-on: ${{ matrix.os }}
+  ## TODO: More minimal test on windows after fixing dev dependency.
```
liamappelbe, line 85: The format we typically use is `TODO(#15):`. Makes it easy to see that every TODO has a github issue

liamappelbe (commented):
Looks good to me, as long as all the TODOs have github issues.

jni_gen/third_party/ApiSummarizer/README.md
```diff
@@ -1,3 +1,46 @@
...
+An early version of ApiSummarizer.
 
+It analyzes java source code / jars and outputs a JSON representation of the public API.
+
+It's currently used in `jni_gen` to get the information of the Java API.
+
+## Build
+When using it via `jni_gen`, the `jni_gen:setup` script will take care of building the jar in appropriate location.
+
+To build the jar manually, run `mvn assembly:assembly` in project root. The jar will be created in `target/` directory.
```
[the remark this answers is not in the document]
mahesh-hegde: No. maven should take care of that.

mahesh-hegde:
Quite interesting bug, before generating test cases I had run google-java-format, which changed the formatting of the mvn-downloaded sources too. That's why when the CI downloads sources from maven and generates bindings, we see mismatch in indentation of doc comments.

mahesh-hegde:
Is there any google repo with GPL as a reference for how to add GPL license to ApiSummarizer files? The FSF website makes it quite long and confusing.

dcharkes:
> Is there any google repo with GPL as a reference for how to add GPL license to ApiSummarizer files? The FSF website makes it quite long and confusing.

I'm not aware of any Google repo.

But, GitHub has a nifty search tool. To find repos that have Java and the GPL license: https://github.com/search?q=stars%3A%3E10000+language%3AJava+license%3Agpl&type=Repositories&ref=advsearch&l=Java&l=

dcharkes:
@mahesh-hegde 

I asked the experts about the licensing.

> Our obligation to make the complete corresponding source code of a program available to end users as open source only triggers if we are distributing compiled software. If we are just publishing source code and are not distributing anything compiled, there should be no copyleft virality concerns.

Currently we are only distributing source, but we might want to distribute precompiled binaries at some point. So, GPLv2 applies.

> The BSD is recognized by GNU/FSF/SFC as a GPLv2-compatible license. This is because compliance with the GPLv2 automatically satisfies compliance with the BSD-3 license, so if I distribute an application that depends on a GPLv2 licensed library, and I distribute the complete corresponding source code for that application, applying the GPLv2 license to the third-party GPLv2 files that I used, and applying the BSD license to all of the other code, this gives end users/the public all of the permissions they need to rebuild that application and to distribute it as a whole under the terms of the GPLv2 license.

So, we can release the Java code under the Dart BSD license. Which also means it doesn't have to live in the third_party folder.

Sorry for asking to change it again, 😅  we're all learning here! 😸 

While you're at updating the folder and license, please also change the java package name to `com.github.dart_lang.jni_gen.apisummarizer` (Note `dart-lang` becomes `dart_lang` as per https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html)

> Looks good to me, as long as all the TODOs have github issues.

Please also update the TODOs to point to GitHub issues.

mahesh-hegde:
Re: package name in java

In the Android package:jni we need a java plugin class to init the JNI. There I used `dev.dart.jni` as package name since dart.dev seems to be official website. Should I update that too?

dcharkes:
> Re: package name in java
> 
> In the Android package:jni we need a java plugin class to init the JNI. There I used `dev.dart.jni` as package name since dart.dev seems to be official website. Should I update that too?

Yes, lets keep it consistent.

And yes, this is a dart-lang package, but it is not part of the Dart SDK. So I'm not sure if `dev.dart` makes sense.

mahesh-hegde:
Added copyright notices to files in jni_gen/.

I am leaving the jni/ folder currently because I have another branch with some changes and don't want extra merge / rebase conflicts.

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/doclet/AnnotationVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.AnnotationValueVisitor;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+
+// AnnotationVisitor mechanism is not perfect right now. There are edge cases which need to be
+// handled but
```
dcharkes, line 16: nit: comment newlines.

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/doclet/AnnotationVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.AnnotationValue;
+import javax.lang.model.element.AnnotationValueVisitor;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeMirror;
+
+// AnnotationVisitor mechanism is not perfect right now. There are edge cases which need to be
```
dcharkes, line 15: nit: This type of comment does not communicate what can be improved. Either document what can be improved in a GitHub issue, or remove the comment.

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/doclet/AstEnv.java
```diff
@@ -0,0 +1,27 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+package com.github.dart_lang.jni_gen.apisummarizer.doclet;
+
+import com.sun.source.util.DocTrees;
+import javax.lang.model.util.Elements;
+import javax.lang.model.util.Types;
+import jdk.javadoc.doclet.DocletEnvironment;
+
+/** Class to hold Utility classes initialized from DocletEnvironment */
```
dcharkes, line 12: Punctuation, no period at the end of sentence.

Also, https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-library-or-type-comments-with-noun-phrases

For example: "Container of utility classes initialized from DocletEnvironment."

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/doclet/SummarizerDocletBase.java
```diff
@@ -0,0 +1,168 @@
...
+
+    @Override
+    public Void visitType(TypeElement e, SummaryCollector collector) {
+      // Supposedly it should visit every type only once.
+      // super.visitType visits the enclosed types and it should do that once.
+      // but its visiting nested classes twice.
+      // This could be a bug in my code, or something related to traversal order.
+      //
+      // Btw, the stack is useless now
+      // TODO: remove stack
```
dcharkes, line 93: Please clean up all TODOs. Either (1) remove the TODO if it is not relevant, or (2) Make a GitHub issue with some background details.

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/doclet/ElementBuilders.java
```diff
@@ -0,0 +1,211 @@
...
+      case CLASS:
+        c.declKind = DeclKind.CLASS;
+        break;
+      case ENUM:
+        c.declKind = DeclKind.ENUM;
+        break;
+      case ANNOTATION_TYPE:
+        c.declKind = DeclKind.ANNOTATION_TYPE;
+        break;
+    }
```
dcharkes, line 38: Consider adding a default case with throwing an exception, otherwise c.declKind will be uninitialized.

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/elements/ClassDecl.java
```diff
@@ -0,0 +1,45 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+package com.github.dart_lang.jni_gen.apisummarizer.elements;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Denotes a class or interface declaration. Here's an example for various kinds of names stored in
```
dcharkes, line 12: https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-library-or-type-comments-with-noun-phrases

```java
/**
 * A class or interface declaration.
 *
 * Here's an example for various kinds of names stored in this structure: simpleName : "Example",
 * binaryName : "dev.dart.sample.Example", parentName : null, packageName : "dev.dart.sample",
 */
```

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/Main.java
```diff
@@ -0,0 +1,205 @@
...
+      cli.addAll(List.of(options.toolOptions.split(" ")));
+    }
+
+    javadoc.getTask(null, fileManager, System.err::println, docletClass, cli, fileObjects).call();
+  }
+
+  public static void main(String[] args) {
+    CommandLine cl = parseArgs(args);
+    config = SummarizerOptions.fromCommandLine(cl);
+    if (config.useAsm) {
```
dcharkes, line 84: Please remove this for now.

jni_gen/test/jackson_core_test/lib/com/fasterxml/jackson/core.dart
```diff
@@ -0,0 +1,4100 @@
+// Autogenerated by jni_gen. DO NOT EDIT!
```
dcharkes, line 1: This file is generated from a Java API with a license, so this generated file must have the same license as a derivative work, and be in a third_party folder.

jni_gen/test/jackson_core_test/generated_files_test.dart
```diff
@@ -0,0 +1,39 @@
...
+const testName = 'jackson_core_test';
+
+void main() async {
+  await generate(isTest: true);
+  test("compare generated bindings for jackson_core", () {
+    compareFiles(testName, 'lib');
+    compareFiles(testName, 'src');
+  });
+  test(
+      'generate and analyze bindings for complete library, '
```
dcharkes, line 22: Sweet! :D

jni_gen/test/jackson_core_test/src/jackson_core_test.c
```diff
@@ -0,0 +1,2188 @@
+#include <stdint.h>
```
dcharkes, line 1: Is this file generated? Then
* add a header similar to the generated file, and
* this file should have the copyright of the file it was generated from.

jni_gen/test/simple_package_test/src/simple_package.c
```diff
@@ -0,0 +1,187 @@
+#include <stdint.h>
```
dcharkes, line 1: Is this file generated? Then add a comment it is generated by JNIgen.

jni_gen/test/test_util/test_util.dart
```diff
@@ -0,0 +1,101 @@
...
+          ),
+          options: options,
+          outputWriter: FilesWriter(
+              cWrapperDir: Uri.directory(src),
+              dartWrappersRoot: Uri.directory(lib),
+              libraryName: testName))
+      .run();
+}
+
+// compare 2 hierarchies, with and without prefix 'test_'
```
dcharkes, line 87: // Compares 2 hierarchies, with and without the prefix 'test_'.

https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

Compare -> Compares

dcharkes (commented):
@mahesh-hegde please address these comments first so that we can merge this PR (instead of tagging new features on to this PR).

As mentioned before:

* Please file github issues for *all* open TODOs.
* All generated files require a copyright/license from the source file.

jni_gen/java/src/main/java/com/github/dart_lang/jni_gen/apisummarizer/disasm/AsmAnnotatedElementVisitor.java
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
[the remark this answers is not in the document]
mahesh-hegde: It's actually working. I will add an example when I add examples to the repo.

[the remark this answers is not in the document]
dcharkes: Can we merge it in a follow up PR with those examples together? That makes this easier to review.

mahesh-hegde:
Yes I am planning to put the examples in follow up PR. But can the disasm
java code stay?

        Mahesh

dcharkes:
> Yes I am planning to put the examples in follow up PR. But can the disasm java code stay?

I think it is bad practise to land code that is not used. What is the problem with adding it in the next PR?

mahesh-hegde:
The --backend asm option calls it. I think it's extra work to remove that
option and corresponding logic in this PR and add back in next.

It actually works right now. I can add a test for that.

        Mahesh

jni_gen/lib/src/bindings/preprocessor.dart
```diff
@@ -0,0 +1,91 @@
...
+        realName = 'ctor';
+      }
+      final sig = method.javaSig;
+      // if method already in super class, assign its number, overriding it.
+      final superNum = superclass?.methodNumsAfterRenaming[sig];
+      if (superNum != null) {
+        // don't rename if superNum == 0
+        final superNumText = superNum == 0 ? '' : '$superNum';
+        // well, unless the method name is a keyword & superNum == 0.
+        // TODO: this logic would better live in a dedicated renamer class.
```
dcharkes, line 71: More TODOs without GitHub issues.

jni_gen/lib/src/bindings/preprocessor.dart
```diff
@@ -0,0 +1,91 @@
...
+        final methodName = superNum == 0 ? kwRename(realName) : realName;
+        method.finalName = '$methodName$superNumText';
+        decl.methodNumsAfterRenaming[sig] = superNum;
+      } else {
+        method.finalName = renameConflict(decl.nameCounts, realName);
+        // TODO: This is too much coupled with renameConflict impl.
```
dcharkes, line 77: More TODOs without GitHub issues.

mahesh-hegde: Fixed.

jni_gen/lib/src/tools/maven_utils.dart
```diff
@@ -0,0 +1,134 @@
...
+        '$i4<maven.compiler.target>$javaVersion</maven.compiler.target>\n'
+        '$i2</properties>\n'
+        '$i4<dependencies>\n'
+        '${depDecls.join("\n")}'
+        '$i2</dependencies>\n'
+        '</project>';
+  }
+}
+
+/// Represents a maven dependency with group ID, artifact ID and version.
```
dcharkes, line 121: https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-library-or-type-comments-with-noun-phrases

/// Maven dependency with group ID, artifact ID and version.

jni_gen/lib/src/util/name_utils.dart
```diff
@@ -0,0 +1,29 @@
...
+  }
+  return [str.substring(0, li), str.substring(li + 1)];
+}
+
+String getLastName(String binaryName) => binaryName.split('.').last;
+
+String getSimpleNameOf(ClassDecl cls) => cls.simpleName;
+
+/// Returns class name as useful in dart.
+/// Eg -> a.b.X.Y -> X_Y
```
dcharkes, line 23: new line in between single line summary and example.

https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

jni_gen/lib/src/util/name_utils.dart
```diff
@@ -0,0 +1,29 @@
...
+String simplifiedClassName(String binaryName) =>
+    getLastName(binaryName).replaceAll('\$', '_');
+
+// Utilities to operate on package names
```
dcharkes, line 27: names -> names.

jni_gen/lib/src/util/rename_conflict.dart
```diff
@@ -0,0 +1,83 @@
...
+    final count = counts[name]!;
+    final renamed = '$name$count';
+    counts[name] = count + 1;
+    return renamed;
+  }
+  counts[name] = 1;
+  return kwRename(name);
+}
+
+// Concats 0 to name if name is a keyword
```
dcharkes, line 16: keyword -> keyword.

dcharkes: Maybe

/// If name is a keyword, postfixes 0.
///
/// Examples:
/// class -> class0
/// foo -> foo

jni_gen/lib/src/writers/files_writer.dart
```diff
@@ -0,0 +1,126 @@
...
+
+import 'package:jni_gen/src/elements/elements.dart';
+import 'package:jni_gen/src/config/wrapper_options.dart';
+import 'package:jni_gen/src/util/find_package.dart';
+
+import 'bindings_writer.dart';
+
+// Writes dart files to specified directory
+// The name of the files themselves is decided by package name
+class FilesWriter extends BindingsWriter {
```
dcharkes, line 17: Maybe

/// Writer of generated dart files. 
///
/// Writes dart files to specified directory.
/// The name of the files themselves is decided by package name.

jni_gen/test/jackson_core_test/generate.dart
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:jni_gen/jni_gen.dart';
+import '../test_util/test_util.dart';
+
+const jacksonPreamble = '// Generated from jackson-core which is licensed under'
```
dcharkes, line 8: 👍

jni_gen/test/jackson_core_test/third_party/src/dartjni.h
```diff
@@ -0,0 +1,172 @@
...
+
+FFI_PLUGIN_EXPORT void SetJNILogging(int level);
+
+FFI_PLUGIN_EXPORT jstring ToJavaString(char *str);
+
+FFI_PLUGIN_EXPORT const char *GetJavaStringChars(jstring jstr);
+
+FFI_PLUGIN_EXPORT void ReleaseJavaStringChars(jstring jstr, const char *buf);
+
+/// For use by jni_gen's generated code
```
dcharkes, line 81: ?

jni_gen/test/jackson_core_test/third_party/src/dartjni.h
```diff
@@ -0,0 +1,172 @@
...
+/// don't use these.
+
+// these 2 fn ptr vars will be defined by generated code library
```
dcharkes, line 84: these -> These

library -> library.

jni_gen/test/jackson_core_test/third_party/src/dartjni.h
```diff
@@ -0,0 +1,172 @@
...
+extern struct jni_context (*context_getter)(void);
+extern JNIEnv *(*env_getter)(void);
+
+// this function will be exported by generated code library
```
dcharkes, line 88: https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

jni_gen/test/simple_package_test/java/dev/dart/simple_package/Example.java
```diff
@@ -0,0 +1,55 @@
+package dev.dart.simple_package;
+
+public class Example {
+  // static final fields
```
dcharkes, line 4: Maybe remove these comments.

Unless you'd like to signify why these language constructs need to stay. (E.g. Static fields for testing JNI bindings to static fields.) But I'd say it is obvious from the context that this a test file that should not be modified to remove the unused things. (We have checked in the expected generated bindings, the test would start failing anyway.)

dcharkes (approved):
A ton of small comments, but no larger concerns anymore.

jni/analysis_options.yaml
```diff
@@ -1,7 +1,7 @@
 include: package:flutter_lints/flutter.yaml
 
 analyzer:
-  exclude: [build/**]
+  exclude: [build/**, third_party/**]
```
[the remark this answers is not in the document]
mahesh-hegde: I had another branch in which ffigen_patch_jni is vendored but not added to git yet. So it will remain in working tree and be analyzed.. It is a stopgap measure to prevent that..

jni_gen/lib/src/bindings/c_bindings.dart
```diff
@@ -0,0 +1,362 @@
...
+  final String _loadEnvCall = '${_indent}load_env();\n';
+
+  String _loadClassCall(String classVar, String internalName) {
+    return '${_indent}load_class_gr(&$classVar, '
+        '"$internalName");\n';
+  }
+
+  String _formalArgs(Method m) {
+    final args = <String>[];
+    // if normal instance method, jobject self
```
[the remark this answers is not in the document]
mahesh-hegde: Removed comment.

In case of constructor, there's no self param but it's actually an instance method. (Impl detail: see https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#NewObject) So hasSelfParam is separate from static / instance check.

jni_gen/lib/src/bindings/dart_bindings.dart
```diff
@@ -0,0 +1,380 @@
...
+  // Name for reference in base class
+  static const _self = 'reference';
+  // symbol lookup function for generated code
+  static const _jlookup = 'jlookup';
+
+  // import prefixes
+  static const ffi = 'ffi.';
+  static const jni = 'jni.';
+
+  // ffi.Pointer<ffi.Void>
```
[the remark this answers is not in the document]
mahesh-hegde: Good catch, I once had FFI import prefix in options that's why it remained as late.

jni_gen/lib/src/bindings/dart_bindings.dart
```diff
@@ -0,0 +1,380 @@
...
+
+    var superName = _jlObject;
+    if (decl.superclass != null) {
+      superName = resolver
+              .resolve((decl.superclass!.type as DeclaredType).binaryName) ??
+          _jlObject;
+    }
+
+    s.write('class $name extends $superName {\n');
+    // fromRef constructor
```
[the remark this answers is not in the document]
mahesh-hegde: Lesson learned. I used to write the layout / section as comments before implementation, in a top-down approach. But forgot to delete the comments.

jni_gen/lib/src/bindings/dart_bindings.dart
```diff
@@ -0,0 +1,380 @@
...
+    final argTypes = [if (hasSelfParam(m)) _voidPtr];
+    for (var param in m.params) {
+      argTypes.add(conv(param.type));
+    }
+    final retType = isCtor(m) ? _voidPtr : conv(m.returnType);
+    return '$retType Function (${argTypes.join(", ")})';
+  }
+
+  // Type for FFI Function signature
+  String dartFfiType(TypeUsage t) {
```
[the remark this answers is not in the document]
mahesh-hegde: I think it can remain along with other methods in binding generator class, because 1) It's not used elsewhere, 2) Future changes may make it require some more context eg: from options.

jni_gen/lib/src/bindings/preprocessor.dart
```diff
@@ -0,0 +1,91 @@
...
+import 'dart:io';
+
+import 'package:jni_gen/src/elements/elements.dart';
+import 'package:jni_gen/src/config/wrapper_options.dart';
+import 'package:jni_gen/src/util/rename_conflict.dart';
+import 'common.dart';
+
+/// preprocessing tasks common to both generators, such as renaming
+/// and exclusion of methods / fields.
+class ApiPreprocessor {
```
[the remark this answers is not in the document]
mahesh-hegde: Again, lesson learned. I used to be sloppy with comments on internal APIs.

jni_gen/lib/src/tools/maven_utils.dart
```diff
@@ -0,0 +1,134 @@
...
+      '-Dmdep.outputFile=$_tempClassPath',
+    ]);
+    final classPathFile = File(_tempClassPath);
+    final classpath = await classPathFile.readAsString();
+    await classPathFile.delete();
+    return classpath;
+  }
+
+  static String _getStubPom(List<MvnDep> deps, {String javaVersion = '11'}) {
+    final i2 = ' ' * 2;
```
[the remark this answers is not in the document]
mahesh-hegde: Doesn't that require string flowing to the left of variable declaration? If so, doesn't look that neat inside a method?

jni_gen/lib/src/util/name_utils.dart
mahesh-hegde: oops I knew this. But it's taking me time to internalize these rules haha.

jni_gen/test/jackson_core_test/third_party/src/dartjni.h
mahesh-hegde: Updated comment. These 2 are function pointer variables initialized by dart code to concrete implementations in `dartjni` base library.

This is because

* we don't know at compile time where in build artifacts folder `dartjni.so` exists -> No compile time linking.
* Default dlopen is RTLD_LOCAL on linux and android at least, thus the symbols are not automatically initialized from previously DLOpen'd library.

I am working around this by having dart code set whatever we need to lookup. I think using `dlsym` in C will become more complex and also probably less portable.

dcharkes: 👍 

I was just confused by the comment, thanks for clarifying.

PR #32, opened 2022-08-26
Add basic config support

mahesh-hegde:
An attempt to add YAML config and simplify total configuration.

* Now maven downloads can be specified in config itself, instead of separately downloading using MavenTools class.
* YAML config with support for overriding values from command line (`-Dproperty.name=value` syntax inspired by Java tools)
* Added test which runs `dart run jni_gen` with equivalent config and compares output.


jni_gen/test/test_data/test_config.yaml
```diff
@@ -0,0 +1,19 @@
+summarizer:
+  use_asm: false
```
dcharkes, line 2: Are we expecting more than two frontends?
If so, make this `frontend` or `parser` with two options `doclet` and `asm`. (And probably the same change with the command line options.)

jni_gen/test/jackson_core_test/jnigen.yaml
```diff
@@ -0,0 +1,36 @@
+maven_downloads:
```
dcharkes, line 1: Please add a section to the README.md with all the yaml config options. (Feel free to copy-paste the setup from package:ffigen's README.md)

jni_gen/lib/src/config/config_provider.dart
```diff
@@ -0,0 +1,132 @@
...
+import 'package:yaml/yaml.dart';
+
+class ConfigError extends Error {
+  ConfigError(this.message);
+  final String message;
+  @override
+  String toString() => message;
+}
+
+class ConfigProvider {
```
dcharkes, line 17: What are the differences between this approach and https://github.com/dart-lang/ffigen/blob/master/lib/src/config_provider/config.dart#L224 ?

It could make sense to use some of the abstractions from FFIgen.

P.S. It is fine to copy code from FFIgen (such as `Specification`), both packages have the same license.

I've filed an issue to make the package FFIgen config_provider more reusable.

jni_gen/lib/src/tools/maven_utils.dart
```diff
@@ -43,7 +43,7 @@ class MvnTools {
   }
 
   /// Create a list of [MvnDep] objects from maven coordinates in string form.
-  static List<MvnDep> makeDependencyList(List<String> depNames) =>
+  static List<MvnDep> deps(List<String> depNames) =>
```
dcharkes, line 46: Rename MvnDep to MavenDependency.

jni_gen/lib/src/config/config_provider.dart
```diff
@@ -0,0 +1,132 @@
...
+  ConfigProvider.of(this.cli, this.yaml);
+  Map<String, String> cli;
+  Map<dynamic, dynamic> yaml;
+
+  static ConfigProvider parseArgs(List<String> args,
```
dcharkes, line 22: Let `parseArgs` return a `Config` object.

Similar to https://github.com/dart-lang/ffigen/blob/master/lib/src/config_provider/config.dart#L21.

This enables us to:

1. catch config errors before doing any code generation
2. write code generation tests by passing in a `Config` we just set the fields of in Dart code instead of through command line args or yaml.

jni_gen/test/config_provider_test.dart
```diff
@@ -0,0 +1,45 @@
...
+
+void main() {
+  final config = ConfigProvider.parseArgs([
+    '--config',
+    join('test', 'test_data', 'test_config.yaml'),
+    '-Dsummarizer.use_asm=true',
+    '--override c_root=src/third_party',
+  ], allowYamlConfig: true);
+  test('getBool', () {
+    expect(config.getBool('summarizer.use_asm'), isTrue);
```
dcharkes, line 17: If using a `Config` object all these test expectations _and_ all usages in the rest of the code become typed. (You could even implement the `operator==` on `Config`, and just check equality in the test here.)

`getString`, `getBool` etc would be private functions that populate this `Config` object.

jni_gen/test/jackson_core_test/jnigen.yaml
dcharkes, line 1: Nice! 🚀

dcharkes (commented):
Nice, I like the `yaml`! (Now I am biased of course by liking FFIgens yamls 😆 )

I think it would be worth making a `Config` object that has all options explicitly. Then all the config options can be accessed in a typed way in the rest of the code base and config can also be easily provided from Dart in unit tests without using yaml or commandline args.

Feel free to take inspiration from https://github.com/dart-lang/ffigen/blob/master/lib/src/config_provider/config.dart

mahesh-hegde:
Few notes:

* `JniGenTask` is basically holding all config, with a run method. In both API based and YAML based entry point. The default constructor is pretty much same options as YAML, although internal representation divides that into 3 components (summarizer, wrapper options, file writer) for current implementation.

* I wrote this config provider so that you can selectively override any YAML property, because sometimes we have to change something for one run. I don't know if ffigen's provider has an equivalent. (I find that quite useful to change eg output directory or SDK path for one command.) That said I can drop this if that's not an important feature.

I once looked at FFIGEN yaml config code and it's quite elaborate and we'll structured. It would indeed be great if we can reuse it.

By the way, it's unfortunate we don't have a json_serializable equivalent for YAML, in which we specify the structs and build_runner generates the boilerplate.

jni_gen/test/test_data/test_config.yaml
mahesh-hegde: Ah I did the change but forgot to change in this yaml.

Btw it is just a test config yaml which is not used for running, only in config provider's unit test. Sorry for the confusion I have to rename it or just use jackson_core_test/jnigen.yaml.

jni_gen/test/jackson_core_test/jnigen.yaml
mahesh-hegde: I thought I'd do that in next PR or so along with runnable examples.

dcharkes:
> `JniGenTask` is basically holding all config

`JniGenTask` is a whole program run, correct? Can it be renamed to `Config` and then the `run` method lives somewhere else and takes the config?

> The default constructor is pretty much same options as YAML

That makes it sound like the yaml parsing method should return that object then.

> although internal representation divides that into 3 components (summarizer, wrapper options, file writer) for current implementation.


In FFIgen, the `Config` also contains info for all stages, and it is passed to all stages.

https://github.com/dart-lang/ffigen/blob/master/lib/src/executables/ffigen.dart#L37

1. Config parsing, and errors if the config contains errors.
2. C parsing takes a config, returns a "library".
3. Code generation takes also the config (well only 1 field) and the library.

E.g. if we were to draw a diagram of how the package works on a high level with the different subpackages and the data flows between those subpackages, it is a simple diagram. (Well actually, you could argue that it is not fully clean, e.g. the "library" classes have the `toBindingString` method and live in the code_gen folder but are instatiated by the parser. It would be slightly cleaner if they would just be data classes flowing from the one subpackage to the other.)

> I wrote this config provider so that you can selectively override any YAML property

What about having a `clone` method on a `Config` object? Is it important to overwrite a yaml property from the command line if we can test with a `Config` object?

mahesh-hegde:
> In FFIgen, the Config also contains info for all stages, and it is passed to all stages.

Currently here it's only WrapperOptions that's passed to binding generators etc.. Rest 2 are individual parts of the pipeline. Yes it's somewhat convoluted.

So what should I do? I think something like:

* Rename `JniGenTask` to `Config`, and add a separate function called `generateJniBindings(Config config)`

* Refactor binding generator parts to take information directly from config. It should suffice to change where `WrapperOptions` is used, since the generate method is supposed to instantiate summarizer and filewriter anyway.

Re: command line overrides. I think we can just have two constructors, one direct YAML and one also parses arguments. Shouldn't be a problem.

I think command line overrides are just convenient, for example you just need to change SDK path or logging level for one run. We don't have to edit yaml for one off uses, or running on a non-standard setup but don't want to change config. (cries in linux distro \#123442).

mahesh-hegde:
Actually that's how it is now - properties in YAML don't have command line
counterparts other than specifying them in -Dfoo.bar=value format.

The only `--` command line option is --config itself, which takes yaml file
as argument.

parseArgs() just takes all args, opens yaml file if one is specified, and
parses rest of property arguments. If you think that's not neat I can
change that to take both a CLI overrides Map and a YAML file explicitly.
But I think that would just create more work in main().

        Mahesh

dcharkes:
> Actually that's how it is now

Neat!

I will have to think a bit about having a unified experience between package:ffigen and package:jnigen. @mannprerak2 how many things do we have that can be both passed by command line and by yaml in package:ffigen? `compiler-opts` anything else?

mannprerak2:
> @mannprerak2 how many things do we have that can be both passed by command line and by yaml in package:ffigen? compiler-opts anything else?

That's it.

mahesh-hegde:
> Error: Bad response: 500 {"message":"Build processing error.","error":true,"url":""}

jni_gen/test/config_provider_test.dart
mahesh-hegde: I decided to go with comparison method in test/ itself, since that will show what was expected and what was obtained.

jni_gen/README.md
```diff
@@ -5,3 +5,27 @@ This enables calling Java code from Dart.
 
 This is a GSoC 2022 project.
 
+Currently this package is highly experimental and proof-of-concept. See `test/jackson_core_test` for an example of generating bindings for a library. It is possible to specify some dependencies to be downloaded automatically through `maven`.
```
dcharkes, line 8: Change `test/jackson_core_test` to a link so it can be clicked.

jni_gen/lib/src/config/config.dart
```diff
@@ -2,7 +2,288 @@
...
+    this.jarOnlyDeps = const [],
+    this.jarDir = defaultMavenJarDir,
+  });
+  List<String> sourceDeps;
+  String sourceDir;
+  List<String> jarOnlyDeps;
+  String jarDir;
+}
+
+/// Configuration for Android SDK sources and stub JAR files.
```
dcharkes, line 39: Nice, this is a very good doc comment! 👌

jni_gen/test/config_test.dart
```diff
@@ -0,0 +1,72 @@
...
+
+const packageTests = 'test';
+final jacksonCoreTests = join(packageTests, 'jackson_core_test');
+final thirdParty = join(jacksonCoreTests, 'third_party');
+final lib = join(thirdParty, 'lib');
+final src = join(thirdParty, 'src');
+final testLib = join(thirdParty, 'test_lib');
+final testSrc = join(thirdParty, 'test_src');
+
+void assertConfigsAreEqual(Config a, Config b) {
```
dcharkes, line 19:
```dart
/// uses `expect` on each field of [a] and [b] to give useful error messages when not equal.
void expectConfigsAreEqual(Config a, Config b) {
```

dcharkes (approved):
lgtm!

Please address the comments and then we merge it! 🚀

mahesh-hegde:
Done. Check once.

jni_gen/lib/src/config/yaml_reader.dart
```diff
@@ -14,19 +14,20 @@ class ConfigError extends Error {
   String toString() => message;
 }
 
-class ConfigProvider {
-  ConfigProvider.of(this.cli, this.yaml);
-  ConfigProvider.fromYaml(this.yaml) : cli = const {};
-  ConfigProvider.fromMap(this.cli) : yaml = const {};
+/// YAML Reader which allows to override specific values from command line.
```
dcharkes, line 17: nit: allows -> enables

To allow is when there's rules that prohibit/disallow things.
To enable is when something was not possible before.

jni_gen/lib/src/tools/maven_tools.dart
```diff
@@ -79,46 +79,45 @@ class MavenTools {
...
+      </${entry.key}>
+      ''');
       }
-      depDecls.add('$i4<dependency>\n'
-          '$i6<groupId>${dep.groupID}</groupId>\n'
-          '$i6<artifactId>${dep.artifactID}</artifactId>\n'
-          '$i6<version>${dep.version}</version>\n'
-          '${otherTags.toString()}\n'
-          '$i4</dependency>\n');
+      depDecls.add('''
```
dcharkes, line 92: Much easier to read :)

jni_gen/test/config_test.dart
```diff
@@ -16,6 +16,8 @@ final src = join(thirdParty, 'src');
 final testLib = join(thirdParty, 'test_lib');
 final testSrc = join(thirdParty, 'test_src');
 
+/// Compares 2 [Config] objects using [expect] to give useful errors when
+/// two fields are not equal.
 void assertConfigsAreEqual(Config a, Config b) {
```
dcharkes, line 21: expectConfigsAreEqual

dcharkes (approved):
A couple of more typos, and then ready to merge. 🎸

PR #35, opened 2022-08-29
Rename java package of simple_test to use com.github.dart_lang.jni_gen prefix instead of dev.dart

mahesh-hegde:


jni_gen/test/bindings_test.dart
```diff
@@ -16,8 +16,8 @@ import 'package:path/path.dart' hide equals;
 import 'package:test/test.dart';
 
 // ignore_for_file: avoid_relative_lib_imports
-import 'simple_package_test/lib/dev/dart/simple_package.dart';
-import 'simple_package_test/lib/dev/dart/pkg2.dart';
+import 'simple_package_test/lib/com/github/dart_lang/jni_gen/simple_package.dart';
```
dcharkes, line 19: Should this be in jni_gen/test/ instead of in the jni_gen root?

dcharkes (approved):

jni_gen/test/bindings_test.dart
mahesh-hegde: It's a relative import. The importing file is in test/

dcharkes: No, I mean the package name.

E.g. conceptually if people ever wanted to import jni_gen.summarizer they would be surprised if there are subpackages pkg2 and simple_package nested in jni_gen. It would be more logical if these packages would be `jni_gen.test.pkg2` and `jni_gen.test.simple_package`. As far as I understand it, we should consider all the java packages in the universe be all in the same packages namespace.

mahesh-hegde:
I don't think packages in test/ hierarchy are importable?

        Mahesh

dcharkes:
Ah okay. (I haven't written Java in a very long time. 😄)

PR #48, opened 2022-09-05
logging

mahesh-hegde:
Last one for today.

### DIFF
--- a/jnigen/lib/src/logging/logging.dart
+++ b/jnigen/lib/src/logging/logging.dart
@@ -1,0 +1,53 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+enum LogLevel {
+  fatal,
+  warning,
+  info,
+  debug,
+  verbose,
+}
+
+class Log {
+  static LogLevel _level = LogLevel.info;
+  static DateTime _began = DateTime.now();
+
+  static String _levelString(LogLevel level) => level.name.toUpperCase();
+
+  static void begin(LogLevel level) {
+    _level = level;
+    _began = DateTime.now();
+  }
+
+  static void _log(LogLevel level, Object? value, {String? color}) {
+    if (level.index <= _level.index) {
+      final duration = DateTime.now().difference(_began);
+      final seconds = (duration.inSeconds % 60).toString().padLeft(2, '0');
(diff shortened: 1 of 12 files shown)


jnigen/lib/src/logging/logging.dart
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+
+enum LogLevel {
```
dcharkes, line 7: We already have these in package logging. Why do we need a duplicate abstraction?

dcharkes (commented):
Do we need a single integration test to see that the logging works as intended? E.g. run our own package with `Process` and capture stdout or stderr to see that it is working as expected.

jnigen/lib/src/logging/logging.dart
mahesh-hegde: I didn't know there was a standard logging package, and defined logging levels + logger. I wrote this class more similar to android's logging levels.

I should probably update this. I will update this PR tomorrow.

Re: unit test, I think it's more effort than worth, we have to update expected output every time we add a logging statement in code.

dcharkes: Yeah, maybe leave the unit test out, unless you for example use the logger to error on the --config path not being found. That we wouldn't change that so often.

jnigen/lib/src/tools/android_sdk_tools.dart
```diff
@@ -68,28 +73,34 @@ task listDependencies(type: Copy) {
...
+          (androidProject == '.') ? '' : ' in $androidProject';
       throw Exception('\n\ngradle exited with exit code ${procRes.exitCode}\n'
           'This can be related to a known issue with gradle. Please run '
-          '`flutter build apk` in $androidProject and try again\n');
+          '`flutter build apk`$inAndroidProject and try again\n');
     }
-    final gradleClassPaths = (procRes.stdout as String).split('\n');
-    if (gradleClassPaths.last.isEmpty) {
-      gradleClassPaths.removeLast();
+    final classpath = (procRes.stdout as String).split('\n');
```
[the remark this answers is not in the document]
mahesh-hegde: conventionally classpath is set of paths. But I agree in some places in jnigen this terminology is inconsistent.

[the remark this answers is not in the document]
mahesh-hegde: nevermind, in this context classpaths makes more sense.

dcharkes (approved):

PR #53, opened 2022-09-07
Jni refactor

mahesh-hegde:
* This PR deletes the old local reference based `JniObject` utilities, and makes `JlObject` [1] the default.
* Also removes dart tool for boilerplate which calls JNI methods, and replaces them by parameterized methods. (although as a downside, these call method / get field utilities depend a lot on casting).
* Exception support (#25) is added to generated code. Currently it works but have to revisit the bindings structure with #52 .
* `JNIEnv` is removed and instead an abstraction called `Indir` is introduced. It's an indirection over actual JNIEnv pointer which manages the thread_local JNIEnv in C code, so that it can be moved across threads in dart code.

A patched version of `ffigen` is vendored in third_party, which is used to generate the bindings in lib/src/third_party.

One change that may be 'controversial' is removing the Jni as singleton type and making JNI management methods static.


liamappelbe (commented):
Indir is a bit of a weird name. Do you expect users to use indir at all, or is it only for internal use? If users are expected to use it, you could just call it env or something, because users don't care that this object is a thin indirection layer on top of JNIEnv. From their perspective it just looks like the env. The fact that it handled threading stuff is an internal detail that should be transparent to users. If it's just for internal use then the name doesn't really matter.

Also, missing copyright boilerplate, as I mentioned in chat.

mahesh-hegde:
The higher level JNI APIs cover method calls and field access.

But other utility methods can only be accessed through Indir. Let's say array related or handling raw references.

My line of thought was that a user accessing such APIs already know how JNI works, and has to refer to JNI docs, so they might be confused where JNIEnv and Indir behaves differently (returning global references, valid across threads).

dcharkes:
Some initial comments:

`Indir` -> `Indirection`? But that is actually a bad name. What is it an indirection for?

> so that it can be moved across threads in dart code.

Sweet!

`JlObject` -> What is "J" "l" ?

> A patched version of ffigen is vendored in third_party, which is used to generate the bindings in lib/src/third_party.

Could we do this in a separate PR?
And check on the CI that the generated file is identical?
And add a README.md to that folder explaining why we need a patched version of ffigen?

> One change that may be 'controversial' is removing the Jni as singleton type and making JNI management methods static.

I'll take a look at the code later.

liamappelbe:
While you're re-adding the copyright boilerplate, I found these files that also seem to be missing it. Can you add the copyright stuff to them before v0.1?

```
jnigen/examples/pdfbox_plugin/example/linux/my_application.cc
jnigen/examples/pdfbox_plugin/example/linux/main.cc
jnigen/examples/pdfbox_plugin/example/linux/my_application.h
jnigen/examples/in_app_java/lib/android_utils/init.dart
jnigen/examples/notification_plugin/lib/init.dart
jnigen/test/test_util/test_util.dart
jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/pkg2/C2.java
jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/simple_package/Example.java
jnigen/test/simple_package_test/lib/init.dart
jni/macos/Classes/jni.c
jni/example/android/app/src/main/java/dev/dart/jni_example/AnyToast.java
jni/example/integration_test/jni_object_test.dart
jni/example/linux/my_application.cc
jni/example/linux/main.cc
jni/example/linux/my_application.h
```

mahesh-hegde:
> What's Indir indirection for?

My bad, the name of the type itself is `JniEnvIndir` but i referred to it as Indir everywhere. I think it makes sense to retain `indir` in variable names, but I will update the external documentation.

> JlObject -> What is "J" "l" ?

Java language, but it might be a bad name. We were removing local reference based JniObject, and I thought using a different name may be wise.

Should I rename it to `JniObject` again? Or something else?

> Explain why patched version of FFIGEN is required.

FFIgen is used to generate `lib/src/third_party/jni_bindings_generated.dart` which binds to Java `jni.h` and our `dartjni.h`. It's a dev dependency and not normal dependency.

I think the user doesn't need to be aware of the fact that a part of library is generated using ffigen. It's just a library for them.

mahesh-hegde:
@liamappelbe

Some of those files are generated as part of default example. I will check the rest of them.

I also think generated code like _init.dart files are too trivial to have a copyright boilerplate on. But prepending preamble to _init.dart is a one-line change, I can do that.

dcharkes:
> > What's Indir indirection for?
> 
> My bad, the name of the type itself is `JniEnvIndir` but i referred to it as Indir everywhere. I think it makes sense to retain `indir` in variable names, but I will update the external documentation.

`JniEnvIndir` Indirect should not be shortened. JniEnv is permissible because it is a type from the JNI header. But here we make our own type. Still I don't like indirect, because _any_ abstraction in programming is indirect. What kind of abstraction is it?

It abstracts the concept of thread-locality of JniEnv away, but we still need to use everywhere where a `JniEnv` would be used.

How about `ThreadSafeJniEnv`?

> > JlObject -> What is "J" "l" ?
> 
> Java language, but it might be a bad name. We were removing local reference based JniObject, and I thought using a different name may be wise.
> 
> Should I rename it to `JniObject` again? Or something else?

`JniObject`, `JavaLanguageObject`, `JavaObject`, `JavaReference`, all of these would be permissible names. Not to be confused with `JObject` which is the actual pointer to C member which holds the Jni Reference which refers to an object in Java memory.

Ideally, we should only ever have a _single_ concept in the API that our user sees. But I need to play around more with the generated code and JNI to see if that is feasible. (We have kind of a similar dichotomy in ObjectiveC bindings where we have the object oriented abstraction and the lower level abstraction.)

Maybe stick with `JniObject` for now.
 
> > Explain why patched version of FFIGEN is required.
> 
> FFIgen is used to generate `lib/src/third_party/jni_bindings_generated.dart` which binds to Java `jni.h` and our `dartjni.h`. It's a dev dependency and not normal dependency.
> 
> I think the user doesn't need to be aware of the fact that a part of library is generated using ffigen. It's just a library for them.

No the user not, so it should not be in the top-level README. But the maintainers probably should be aware. (And I am not going to remember likely, so I will ask again in 3 months if we don't put it in a readme next to the folder.)

mahesh-hegde:
> `ThreadSafeJniEnv`

Right. One alternative I thought of is `GlobalJniEnv`.

So a short summary of renames & changes to be done

* `JniEnvIndir` -> `GlobalJniEnv`, `indir` -> `env`

* `JlObject` -> `JniObject`, `JlClass` -> `JniClass`, `JlString` -> `JniString`

* Put a README in FFIgen one-use patch.

jni/src/global_jni_env.c
```diff
@@ -0,0 +1,1188 @@
+// Copyright (c) 2022, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+#include "global_jni_env.h"
+jint Jei_GetVersion() {
+    attach_thread();
+    return (*jniEnv)->GetVersion(jniEnv);
```
dcharkes, line 8: What is the conceptual relationship between dartjni.c and global_jni_env.c ?

jniEnv lives in dartjni.c while globalEnv abstracts over that.

Does the generated code talk with both? If yes, then it is not a full abstraction and maybe these two files should just be merged.

jni/lib/src/third_party/jni_bindings_generated.dart
```diff
@@ -1489,1615 +1526,2193 @@ class JNINativeInterface extends ffi.Struct {
...
+      GetDirectBufferAddress;
+
+  external ffi.Pointer<ffi.NativeFunction<JLong Function(JObject)>>
+      GetDirectBufferCapacity;
+
+  external ffi.Pointer<ffi.NativeFunction<ffi.Int32 Function(JObject)>>
+      GetObjectRefType;
+}
+
+extension GlobalJniEnvExtension on ffi.Pointer<GlobalJniEnv> {
```
dcharkes, line 2415: Nice, I like this, makes interacting much nicer.

jni/lib/src/third_party/jni_bindings_generated.dart
```diff
@@ -1489,1615 +1526,2193 @@ class JNINativeInterface extends ffi.Struct {
...
+  @pragma('vm:prefer-inline')
```
dcharkes, line 2416: The VM should be smart enough to make its own inlining decisions. Unless you have measured this to have a performance impact, please remove. (If you have measured this to make a performance impact, file a bug with a reproduction and we should fix it.)

jni/example/lib/main.dart
```diff
@@ -9,104 +9,102 @@ import 'package:flutter/material.dart';
...
-int random(int n) {
+int randomUsingEnv(int n) {
   final arena = Arena();
-  final jniEnv = jni.getEnv();
-  final randomCls = jniEnv.FindClass("java/util/Random".toNativeChars(arena));
-  final ctor = jniEnv.GetMethodID(
+  final env = Jni.env;
+  final randomCls = env.FindClass("java/util/Random".toNativeChars(arena));
+  final ctor = env.GetMethodID(
       randomCls, "<init>".toNativeChars(arena), "()V".toNativeChars(arena));
```
dcharkes, line 38: How are the method signatures constructed?

Can we construct `"()V"`, `"(I)I"`, `"()D"`, and ones with parameter types from a helper data structure? Would that be useful to have for users directly interacting with package:jni, instead of jnigen? (I presume we generate these in jnigen.) Feel free to file a new issue for this rather than address this in this PR.

jni/ffigen.yaml
```diff
@@ -73,7 +82,7 @@ typedefs:
     'jvalue': 'JValue'
 preamble: |
   // Autogenerated file. Do not edit.
-  // Generated from an annotated version of jni.h provided in Android NDK
+  // Generated from an annotated version of jni.h provided in Android NDK.
```
dcharkes, line 85: Nice catch!

dcharkes (commented):
Some initial notes from playing with the code.

jnigen/examples/pdfbox_plugin/lib/third_party/_init.dart
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import "dart:ffi";
+import "package:jni/internal_helpers_for_jnigen.dart";
+
+final Pointer<T> Function<T extends NativeType>(String sym) jlookup =
```
dcharkes, line 22: `jlookup` -> `lookup`/`javaLookup`?

jni/src/global_jni_env.h
```diff
@@ -0,0 +1,206 @@
...
+    void (*SetObjectField)(jobject obj, jfieldID fieldID, jobject val);
+    void (*SetBooleanField)(jobject obj, jfieldID fieldID, jboolean val);
+    void (*SetByteField)(jobject obj, jfieldID fieldID, jbyte val);
+    void (*SetCharField)(jobject obj, jfieldID fieldID, jchar val);
+    void (*SetShortField)(jobject obj, jfieldID fieldID, jshort val);
+    void (*SetIntField)(jobject obj, jfieldID fieldID, jint val);
+    void (*SetLongField)(jobject obj, jfieldID fieldID, jlong val);
+    void (*SetFloatField)(jobject obj, jfieldID fieldID, jfloat val);
+    void (*SetDoubleField)(jobject obj, jfieldID fieldID, jdouble val);
+    jmethodID (*GetStaticMethodID)(jclass clazz, const char * name, const char * sig);
```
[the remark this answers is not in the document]
mahesh-hegde: There's actually one for each method in JNI, and even JNI headers don't document these functions. So the user using these is supposed to know the C equivalent IG.

jni/src/global_jni_env.c
mahesh-hegde: JNIEnv lives in jni.h from NDK actually. `dartjni.c` provides other core infrastructure specific to dart. global_jni_env is just a wrapper over env with handful of methods.

True. The orgnization is clumsy and I should revisit it.

jnigen/lib/src/bindings/c_bindings.dart
```diff
@@ -345,12 +349,12 @@ class CPreludes {
       '#include "dartjni.h"\n'
       '\n';
   static const defines = 'thread_local JNIEnv *jniEnv;\n'
-      'struct jni_context jni;\n\n'
-      'struct jni_context (*context_getter)(void);\n'
+      'struct JniContext jni;\n\n'
```
dcharkes, line 352: I think struct can be omitted because we have 

```c
typedef struct JniContext {
	JavaVM *jvm;
	jobject classLoader;
	jmethodID loadClassMethod;
	jobject currentActivity;
	jobject appContext;
} JniContext;
```

jnigen/lib/src/bindings/c_bindings.dart
```diff
@@ -345,12 +349,12 @@ class CPreludes {
...
+      'struct JniContext (*context_getter)(void);\n'
       'JNIEnv *(*env_getter)(void);\n'
       '\n';
   static const initializers =
-      'void setJniGetters(struct jni_context (*cg)(void),\n'
+      'void setJniGetters(struct JniContext (*cg)(void),\n'
```
dcharkes, line 357: struct can be omitted

jni/src/global_jni_env.h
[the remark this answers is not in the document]
dcharkes: Hm, we should at least refer to the official documentation then somewhere.

And we should consider documenting what we reverse engineered ourselves as about who owns the arguments and what lifetime they have. Otherwise every time we need to reverse engineer again (e.g. I'm going to forget in 3 months.) and any potential users of our generated Dart file is going to ask the same questions.

jnigen/test/jackson_core_test/third_party/lib/_init.dart
```diff
@@ -0,0 +1,22 @@
+// Generated from jackson-core which is licensed under the Apache License 2.0.
```
dcharkes, line 1: This file actually doesn't contain any code from the original header file, so it doesn't necessarily need the same copyright. But since it is generated, maybe it should just have the identical preamble to the other files.

Or is it not generated?

jni/example/lib/main.dart
mahesh-hegde: I replied to the issue.

jnigen/examples/pdfbox_plugin/lib/third_party/_init.dart
mahesh-hegde: lookup is ruled out, because it might conflict with some variable.

`javaLookup` maybe.

jnigen/lib/src/bindings/c_bindings.dart
mahesh-hegde: Nice catch.

jnigen/test/jackson_core_test/third_party/lib/_init.dart
mahesh-hegde: Generated.

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,543 @@ import 'dart:ffi';
...
+  /// This implementation looks up the method and calls [callStaticMethod].
+  T callStaticMethodByName<T>(String name, String signature, List<dynamic> args,
+      [int? callType]) {
+    final id = getStaticMethodID(name, signature);
+    return callStaticMethod<T>(id, args, callType);
+  }
+}
+
+/// A high level wrapper over a JNI class reference.
+class JniClass extends JniReference implements Finalizable {
```
dcharkes, line 395: JniClass should extends JniObject. In the JNI header, jclass extends jobject.

dcharkes (commented):
### `NativeFinalizer` feedback

This looks pretty sweet! Some high level comments:

* `JniClass` should extends `JniObject` as jclass extends jobject in the jni interface.
* `JniObject` should not have an extra class reference, it should just return a JniClass and the class methods should be on that object.
* The finalizer, constructor attaching the finalizer, and `delete` method can be in the `JniReference` class.

Otherwise from inspecting the code, the code looks good. 👌 🔥

jni/example/lib/main.dart
```diff
@@ -9,104 +9,99 @@ import 'package:flutter/material.dart';
...
+      final i = arena<JValue>();
+      i.ref.i = n;
+      final res = env.CallStaticObjectMethodA(cls, mId, i);
+      final str = env.asDartString(res);
+      env.deleteAllRefs([res, cls]);
+      return str;
+    });
+
+int randomUsingEnv(int n) => using((arena) {
+      final arena = Arena();
```
dcharkes, line 32: Stray line. (Causing your Strings to _not_ be deleted! 😆)

jni/test/jni_test.dart
```diff
@@ -22,26 +20,35 @@ void main() {
...
-    final env = jni.getEnv();
-    expect(env.GetVersion(), isNot(equals(0)));
+    expect(Jni.env.GetVersion(), isNot(equals(0)));
   });
 
   test('Manually lookup & call Long.toHexString static method', () {
     // create an arena for allocating anything native
     // it's convenient way to release all natively allocated strings
     // and values at once.
     final arena = Arena();
```
dcharkes, line 51: Convert tests to `using((Arena arena)` as well.

jni/lib/src/jni.dart
```diff
@@ -188,120 +139,136 @@ class Jni {
...
-    return _bindings.GetClassLoader();
-  }
+  static JObject getApplicationClassLoader() => _bindings.GetClassLoader();
 
   /// Returns class reference found through system-specific mechanism
-  JClass findClass(String qualifiedName) {
+  static JClass findClass(String qualifiedName) {
     final nameChars = qualifiedName.toNativeChars();
     final cls = _bindings.LoadClass(nameChars);
     calloc.free(nameChars);
```
dcharkes, line 182: Use Arena.

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,543 @@ import 'dart:ffi';
...
+      break;
+    default:
+      throw UnsupportedError('Unknown type $T');
+  }
+  return result;
+}
+
+T _callMethod<T>(
+    int? callType, List<dynamic> args, Function(int, Pointer<JValue>) f) {
+  final jArgs = JValueArgs(args);
```
dcharkes, line 192: Use Arena instead.

If `_callOrGet` can throw (or we change it in the future to throw) then the `free` two lines below doesn't happen).

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,543 @@ import 'dart:ffi';
...
+  T callStaticMethodByName<T>(String name, String signature, List<dynamic> args,
+      [int? callType]) {
+    final id = getStaticMethodID(name, signature);
+    return callStaticMethod<T>(id, args, callType);
+  }
+
+  /// Create a new instance of this class with [ctor] and [args].
+  JniObject newObject(JMethodID ctor, List<dynamic> args) {
+    _assertNotDeleted();
+    final jArgs = JValueArgs(args);
```
dcharkes, line 500: Use Arena

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,543 @@ import 'dart:ffi';
...
-  final JObject _obj;
-  final JClass _cls;
-  bool _deleted = false;
-  JniGlobalObjectRef._(this._obj, this._cls);
+class JniString extends JniObject {
+  /// Construct a new [JniString] with [reference] as its underlying reference.
+  JniString.fromRef(JString reference) : super.fromRef(reference);
+
+  static JString _toJavaString(String s) {
+    final chars = s.toNativeUtf8().cast<Char>();
```
dcharkes, line 516: Use Arena. (For if NewStringUTF would ever throw.)

jni/lib/src/jni.dart
```diff
@@ -46,102 +41,58 @@ DynamicLibrary _loadJniHelpersLibrary(
...
-    inst._bindings.SetJNILogging(logLevel);
     final jArgs = _createVMArgs(
       options: jvmOptions,
       classPath: classPath,
       version: jniVersion,
       ignoreUnrecognized: ignoreUnrecognized,
     );
-    inst._bindings.SpawnJvm(jArgs);
+    _bindings.SpawnJvm(jArgs);
     _freeVMArgs(jArgs);
```
dcharkes, line 95: Use a single arena to allocate everything, with a using wrapping the whole spawn body. Then the whole `_freeVMArgs` isn't needed.

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,543 @@ import 'dart:ffi';
...
-  @pragma('vm:prefer-inline')
-  void _checkDeleted() {
+  /// Reference to class of the reference obtained using `GetObjectClass`. It's
+  /// populated only if any methods called on this object require it.
+  ///
+  /// This is nullable field instead of a late field, because we need to check
+  /// if it's populated and delete if so.
+  JObject? _classRef;
+
+  JObject get _class {
```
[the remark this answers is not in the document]
mahesh-hegde: Actually most reflective methods require class reference. Eg callMethodByName() or getFieldId().

So `obj.callMethodByName()` becomes either
`obj.callMethod(obj.getClass().getMethod(...)` or obj requires to have a late initialized `JniClass` member instead.

jni/lib/src/env_extensions.dart
```diff
@@ -37,7 +23,7 @@ extension AdditionalJniEnvMethods on Pointer<JniEnv> {
     final result = chars.cast<Utf8>().toDartString();
     ReleaseStringUTFChars(jstring, chars);
     if (deleteOriginal) {
-      DeleteLocalRef(jstring);
+      DeleteGlobalRef(jstring);
     }
     return result;
   }
```
dcharkes, line 29: `asJString` use arena.

dcharkes (commented):
### `Arena` comments

The use of `Arena` and `using` is great! It covers the exceptional control flow and we can't forget to release things. 👍 

I found some more places in the code where `free` is still called manually, please fix those as well.

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,543 @@ import 'dart:ffi';
...
+  JniObject.fromRef(this.reference) {
+    _finalizer.attach(this, reference, detach: this);
+  }
+
+  /// Pass the underlying reference into [function] and return the result.
+  ///
+  /// Result of [function] takes ownership of [reference], and this object is
+  /// marked as deleted. [function] can be constructor of a subclass of JniClass
+  /// in which case this function will be useful to cast an expression.
+  T into<T extends JniObject>(T Function(JObject) function) {
```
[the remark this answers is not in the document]
mahesh-hegde: It's supposed to be for cast between JniObject subclasses. I can't figure out how to implement something like `C1_Instance.cast<Class2>()`.

jni/lib/src/jni_object.dart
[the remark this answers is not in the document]
dcharkes: > `obj.callMethod(obj.getClass().getMethod(name, sig), args)`

Ah that would be bad. I had only seen `getFieldID` which doesn't need an object, only the class.

> obj requires to have a late initialized `JniClass` member instead.

Yes, that is a better solution. (Since `JniObject` and `JniClass` are in the same file, JniObject should have access to JniClass internal _reference.)

jni/lib/src/jni_object.dart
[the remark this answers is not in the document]
dcharkes: Lets file a GitHub issue to support those casts and delete the method for now. Then we also don't have to add `_disownReference` and our resource management will be super clean.

jni/lib/src/jni_object.dart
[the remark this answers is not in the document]
mahesh-hegde: Last question. Having late initialized JniClass means it won't be deleted in eg: an arena. So now we have to make it nullable, check if initialized and delete it in JniObject's delete() method. Doesn't that mean we are almost back to square one?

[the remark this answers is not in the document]
dcharkes: > So now we have to make it nullable, check if initialized and delete it in JniObject's delete() method.

Correct.

> Doesn't that mean we are almost back to square one?

Yes and no. Yes, we have a check. No, if we forgot to delete it, the NativeFinalizer would delete it.

I think that is the cleanest solution. (I'll try to think some more about what kind of patterns would be nice with kind of coupled native resources.) This is really an artifact of the JNI API where you have to pass the object _and_ the class for those methods. And we want to provide something nicer for our users.

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,498 @@ import 'dart:ffi';
...
+extension JniReferenceUseExtension<T extends JniReference> on T {
+  /// Applies [callback] on [this] object and then delete the underlying JNI
+  /// reference, returning the result of [callback].
+  R use<R>(R Function(T) callback) {
+    _ensureNotDeleted();
+    try {
+      final result = callback(this);
+      delete();
+      return result;
+    } catch (e) {
```
dcharkes, line 67:
```dart
    try {
      return callback(this);
    } finally {
      delete();
    }
```

jni/lib/src/jni_object.dart
```diff
@@ -7,178 +7,498 @@ import 'dart:ffi';
...
+  JniObject.fromRef(JObject reference) : super.fromRef(reference);
+
+  JniClass? _jniClass;
 
-  /// Returns underlying [JObject] of this [JniObject].
-  JObject get jobject {
-    _checkDeleted();
-    return _obj;
+  JniClass get _class {
+    _jniClass ??= getClass();
```
dcharkes, line 246: `return _jniClass ??= getClass();`

jni/example/README.md
```diff
@@ -1,16 +1,3 @@
 # jni_example
-
 Demonstrates how to use the jni plugin.
```
dcharkes, line 2: the pubspec.lock in this folder is out of date:

```yaml
  jni:
    dependency: "direct main"
    description:
      path: ".."
      relative: true
    source: path
    version: "0.0.1"
```

jni/lib/src/env_extensions.dart
```diff
@@ -37,23 +23,23 @@ extension AdditionalJniEnvMethods on Pointer<JniEnv> {
...
 
   /// Return a new [JString] from contents of [s].
-  JString asJString(String s) {
-    final utf = s.toNativeUtf8().cast<Char>();
-    final result = NewStringUTF(utf);
-    malloc.free(utf);
-    return result;
-  }
+  JString asJString(String s) => using((arena) {
```
dcharkes, line 32:
```dart
  JString asJString(String s) => using((arena) {
        final utf = s.toNativeUtf8(allocator: arena).cast<Char>();
        return NewStringUTF(utf);
      });
```

dcharkes (approved):
lgtm

PR #59, opened 2022-09-20
Windows support

mahesh-hegde:
The first steps of windows support for generated code.

Need to set path to $JAVA_HOME\bin\server. So in practice it's only useful now for testing & development on a windows machine.

### DIFF
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -210,6 +210,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
+      - run: git config --global core.autocrlf true
       - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
       - run: dart pub get
       - run: dart test -j 1
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -89,6 +89,7 @@ jobs:
           flag-name: jnigen_tests
           parallel: true
           path-to-lcov: ./jnigen/coverage/lcov.info
+        if: ${{ matrix.sdk == 'stable' }}
 
   test_summarizer:
     runs-on: ubuntu-latest
@@ -191,7 +192,7 @@ jobs:
       - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
       - run: dart pub get
       - run: dart run jni:setup
-      - run: dart test -j 1
+      - run: dart test
 
   test_jnigen_windows_minimal:
     needs: [analyze_jnigen]
@@ -213,7 +214,7 @@ jobs:
       - run: git config --global core.autocrlf true
       - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
       - run: dart pub get
-      - run: dart test -j 1
+      - run: dart test
 
   build_jni_example_linux:
     runs-on: ubuntu-latest
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -188,8 +188,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
-      - run: |
-          "${env:JAVA_HOME}bin/server" >> $GITHUB_PATH
+      - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
       - run: dart pub get
       - run: dart run jni:setup
       - run: dart test -j 1
@@ -211,8 +210,7 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
-      - run: |
-          "${env:JAVA_HOME}bin/server" >> $GITHUB_PATH
+      - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
       - run: dart pub get
       - run: dart test -j 1
 
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -214,6 +214,7 @@ jobs:
       - run: git config --global core.autocrlf true
       - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
       - run: dart pub get
+      - run: dart run jnigen:setup
       - run: dart test
 
   build_jni_example_linux:
--- a/.github/workflows/test-package.yml
+++ b/.github/workflows/test-package.yml
@@ -188,7 +188,8 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '11'
-      - run: ${env:JAVA_HOME}bin/server >> $GITHUB_PATH
+      - run: |
+          "${env:JAVA_HOME}bin/server" >> $GITHUB_PATH
       - run: dart pub get
       - run: dart run jni:setup
       - run: dart test -j 1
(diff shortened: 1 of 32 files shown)


mahesh-hegde:
Ah CRLF headache.

.github/workflows/test-package.yml
```diff
@@ -157,6 +169,52 @@ jobs:
...
+          cache: true
+          cache-key: 'flutter-:os:-:channel:-:version:-:arch:-:hash:'
+      - uses: actions/setup-java@v2
+        with:
+          distribution: 'zulu'
+          java-version: '11'
+      - run: Add-Content $env:GITHUB_PATH "$env:JAVA_HOME\bin\server"
+      - run: dart pub get
+      - run: dart run jni:setup
+      - run: dart test -j 1
```
dcharkes, line 194: Why does it fail without `-j 1`? Is our API not concurrency-safe?

Can we try to figure out what concurrency guarantees we can and cannot make? (Also, why doesn't it fail on Linux?)

jni/README.md
```diff
@@ -16,6 +16,9 @@ This is intended for one-off / debugging uses of JNI, as well as providing a bas
 
 __To generate type-safe bindings from Java libraries, use `jnigen`.__
 
+## SDK Note
+Dart standalone is supported, but due to some current limitations of the `pubspec` format, `dart` command must be from Flutter SDK and not dart SDK. See dart-lang/pub#3563.
```
dcharkes, line 20: Use a markdown link to add the full URL so people can click it in non-GitHub markdown editors as well. (Or copy the link if reading in txt format.)

jnigen/lib/src/generate_bindings.dart
```diff
@@ -31,12 +31,12 @@ Future<void> generateJniBindings(Config config) async {
   final mavenDl = config.mavenDownloads;
   if (mavenDl != null) {
     final sourcePath = mavenDl.sourceDir;
-    Directory(sourcePath).create(recursive: true);
+    await Directory(sourcePath).create(recursive: true);
```
dcharkes, line 34: Nice catch, this should have created crashes earlier.

dcharkes (approved):
Nice!

.github/workflows/test-package.yml
mahesh-hegde: The creation and deletion of some stub files conflicts obviously. Windows seems just more brittle so errors were caught.

It can be countered by attaching timestamp to every temp file. But is that worth the effort?

jnigen/lib/src/generate_bindings.dart
mahesh-hegde: unawaited_futures should be default lint rule or something.

.github/workflows/test-package.yml
dcharkes: > some stub files conflicts

Is that files that we create? Or are these created by `dart run`? I'd like to understand the issue.

mahesh-hegde: Yes for example __temp_pom.xml.

dcharkes: > It can be countered by attaching timestamp to every temp file. 

I think it is more common to use https://api.dart.dev/stable/2.18.1/dart-io/Directory/createTemp.html.

And making tests conflict with each other is a bug in our tests that we should fix. (Feel free to fix either in this PR or to file an issue and make a separate PR.)

mahesh-hegde:
Done.

Although I think tests need to be streamlined better in future.

dcharkes:
> Although I think tests need to be streamlined better in future.

Feel free to file a GitHub issue with a bullet list of improvements.

jni/bin/setup.dart
```diff
@@ -154,7 +154,23 @@ void main(List<String> arguments) async {
   } else {
     // pass srcDir absolute path because it will be passed to CMake as arg
     // which will be running in different directory
-    await build(options, srcDir.absolute.path, buildDir.path);
+    final now = DateTime.now().microsecondsSinceEpoch;
+    final tempDirUri = Uri.directory(".dart_tool/cmake-build-$now");
```
dcharkes, line 158: Why not use the actual temp folder from the OS? Is there a need to have a folder in `.dart_tool`?

Also, `DateTime.now().microsecondsSinceEpoch` is not guaranteed to be unique. https://api.dart.dev/stable/2.18.1/dart-io/Directory/createTemp.html is guaranteed to be unique.

dcharkes (approved):
lgtm with one more comment.

Thanks @mahesh-hegde! 🎸

jni/bin/setup.dart
mahesh-hegde: It's a small change but before doing that, is there any security best practice I should be aware of before using Directory.systemTemp?

dcharkes: No, it's just a directory with a unique name somewhere on disk.

I suggest using `createTemp` and using a prefix for what you're doing. If you use `Directory.systemTemp` you have to invent a unique name yourself.

mahesh-hegde: Are you suggesting

```
Directory.systemTemp.createTemp("dartjni_native_build")
```

Or 

```
Directory(".dart_tool").createTemp("cmake_native_build")
```

?

dcharkes: Ah, I misunderstood. The first one. (Somehow in my brain createTemp was an abstract method that always used systemTemp under the hood.) Thanks for clarifying!

dcharkes (approved):
Should we make it v0.1.1 with a changelog entry? So that we can publish v0.1.1 with Windows support?

mahesh-hegde:
This windows support is more for myself than normal user.

In CI there's a line which adds JAVA_HOME\bin\server to PATH. So this is useful to run tests for generated bindings on a windows development machine. But I don't think we can call it "windows is supported".

dcharkes:
Well, if the jni and jnigen tests pass, and users can spawn a JVM on a windows host, then we support it right?

Or do we only support running the generator on Windows and then running the generated code on Android devices? (It looks like from the CI that we actually support spawning a JVM, because we need to run the AST summarizer.)

In other words, users could play around with it. Would you want to call it experimental support then? (The package is already experimental, so maybe that doesn't make sense. 😆 )

What do we need to add for full Windows support? (Do we have a GitHub issue?)

mahesh-hegde:
I think the big issue is the need to manually set the PATH. Should I just add that to README?

dcharkes:
> I think the big issue is the need to manually set the PATH. Should I just add that to README?

Yeah, that sounds good.

mahesh-hegde:
> FileSystemException: Cannot rename file to 'build\jni_libs\dartjni.dll', path = 'C:\Users\RUNNER~1\AppData\Local\Temp\dartjni_native_build_8aaa001a\Debug\dartjni.dll' (OS Error: The system cannot move the file to a different disk drive.

Windows moment.

dcharkes:
> > I think the big issue is the need to manually set the PATH. Should I just add that to README?
> 
> Yeah, that sounds good.

Should we make it v0.1.1 and close https://github.com/dart-lang/jnigen/issues/15 ? Or are there still open issues for Windows support that need to be addressed later?

mahesh-hegde:
I don't recall any other issues.

        Mahesh

dcharkes:
Please bump the version to v0.1.1 and add Windows support to the changelog. Then I will release a new version after merging. 🎸

dcharkes:
CI is failing.

```
[ApiSummarizer] Error: Could not find or load main class com.github.dart_lang.jnigen.apisummarizer.Main
[ApiSummarizer] Caused by: java.lang.ClassNotFoundException: com.github.dart_lang.jnigen.apisummarizer.Main
Fatal: Cannot parse summary: FormatException: Unexpected end of input (at character 1)
```

dcharkes:
🚀 😄

Issue #84, opened 2022-10-02
clang-format generated C bindings


mahesh-hegde:
It's a good idea if generated bindings are formatted consistently.

In theory we can run `clang-format` on generated .c file once in the generator.

In practice user may open the file in an editor with auto format (Eg VSCode with C++ plugin). Then it will reformat the file. (Yes user should not edit the generated file. But one might open the file and press Ctrl+S out of habit, which will reformat.)

Is it a good idea to copy the `.clang-format` file from `jni` package into generated c bindings folder, and format according to that?

mahesh-hegde:
@dcharkes

#### Approach 1: Do not format generated bindings
++ No requirement of clang-format tool
-- No consistent formatting

#### Approach 2: Format using default config
++ No config file required
-- Mismatch with config used by dart SDK and package:jni
-- Requires clang-format on user's end.

#### Approach 3: Copy .clang-format from `jni` into generated code directory.
++ Consistent with other packages
-- Requires clang-format on user's end.

dcharkes:
How about we try if we find `clang-format` and then use it, if we don't find it we output a warning but still continue.

Since we generate the code we could dictate the formatting. Do we need to copy the `.clang-format` file in, or can we pass clang-format a config file with an argument?

dcharkes:
Yeah that's true.

I wouldn't mind having a config file. I think it is good practise to specify the way code should be formatted, such file does exactly that.

If anyone later on doesn't like it, we can always add a config option to not generate the the formatting file.

PR #87, opened 2022-10-05
Few improvements in generated C code including exception support.

mahesh-hegde:
* C bindings no longer use fully qualified names. Instead, it uses number-renamed simple names. The classes are sorted in the summarizer to ensure the order is deterministic. This enables shorter symbol names and less verbose bindings.

* C bindings are formatted using same style as Dart SDK, using clang-format. If clang-format is not available, jnigen will issue a warning.
Closes: #84

* C bindings return `JniResult`. The older `Jni.checkException` stopgap is removed. It reduces one line per function binding in dart. Lastly exceptions will work properly on android
Closes: #56

* ~~Added `root_package` option, which allows simpler hierarchy if all classes are from same package. For example, instead of `org/apache/pdfbox/pdmodel.dart` we can create `pdmodel.dart`, by telling that root_package is `org.apache.pdfbox`.~~


jnigen/example/pdfbox_plugin/jnigen.yaml
```diff
@@ -33,7 +33,15 @@ c_root: 'src/'
...
-dart_root: 'lib/third_party/'
+dart_root: 'lib/src/third_party/'
+
+## Root package configuration, this allows to write bindings to $dart_root/pdmodel.dart instead of
+## $dart_root/org/apache/pdfbox/pdmodel.dart assuming the org.apache.pdfbox is common to all subpackages
+## included in this config.
+##
+## Since every package maps to dart file, even if all your classes are in one package A.B.C, root package
+## will be A.B and not A.B.C
+root_package: org.apache.pdfbox
```
dcharkes, line 44: Maybe this should be called something else.

The goal is to avoid unnecessarily deep nesting (which is _not_ avoided in Java itself). (I don't like the Java approach either, but is unambiguous. Maybe we should keep the nesting?)

This is slightly related to output folder in a sense `dart_root`. (Should we rename `dart_root` to something like `output` Since it is just a path, and we can reason about it on the file/folder level. Or even 

```yaml
output:
  dart: path/where/dart/files/are/generated.dart
  c: path/where/c/file/is/generated.c
```
)

```yaml
output:
  dart:
    path: 'path/where/dart/files/are/generated.dart'
    single_file: true
    un_nest: 'org.apache.pdfbox'
  c: 'path/where/c/file/is/generated.c'
```

jnigen/example/pdfbox_plugin/lib/pdfbox_plugin.dart
```diff
@@ -0,0 +1,5 @@
+/// File merely exporting the generated bindings from lib/src/third_party
+library pdfbox_plugin;
+
+export 'src/third_party/pdmodel.dart';
```
dcharkes, line 4: Should we generate a `src/third_party/_all.dart` from jnigen?

That way we don't have to deal with manually adding extra exports here when we add classes.

jni/lib/src/jni.dart
```diff
@@ -256,3 +255,54 @@ extension ProtectedJniExtensions on Jni {
     return lookup;
   }
 }
+
+extension AdditionalEnvMethods on Pointer<GlobalJniEnv> {
```
dcharkes, line 259: (The GitHub interface is not good at doing showing diffs, please only move around things if really needed.)

dcharkes (commented):
First look

dcharkes:
> * Added `root_package` option, which allows simpler hierarchy if all classes are from same package. For example, instead of `org/apache/pdfbox/pdmodel.dart` we can create `pdmodel.dart`, by telling that root_package is `org.apache.pdfbox`.

I presume we add this because it is annoying for people. 😄 
Though, Java explicitly doesn't support this.
The question is should we make our bindings look mostly like Java or mostly like Dart? (With ffigen in C, we say that when you write code against generated bindings, you're writing C in Dart syntax. So we're okay with violating dart naming for identifiers etc. With that reasoning, maybe we should stick with the nested folders, because it is more Java-esque.)

I would like to unify everything that influences _only_ output in the config:

```yaml
output:
  dart:
    path: 'lib/src/third_party/pdfbox/'
    output_structure: single_file || package_structure
    un_nest: 'org.apache.pdfbox'
  c:
    ## Root for generated C bindings.
    root: 'src/'
    ## C files can be stored in a different sub-directory inside c_root.
    ##
    ## We have a guideline to keep all generated code in third_party/ since original
    ## project's license applies to generated code. So we specify third_party/ as
    ## c_subdir while keeping generated CMakeLists.txt in src/.
    subdir: 'third_party/'
```

Also, maybe we could clean up the C output

```yaml
output:
  dart:
    path: 'lib/src/third_party/pdfbox/'
    output_structure: single_file || package_structure
    un_nest: 'org.apache.pdfbox' # Undecided if we should do this.
  c:
    path: 'src/third_party/' # both generated C file and makefile?
    cmake_path: 'src/' # To make Flutter happy?
```

dcharkes:
> * C bindings no longer use fully qualified names. Instead, it uses number-renamed simple names. The classes are sorted in the summarizer to ensure the order is deterministic. This enables shorter symbol names and less verbose bindings.

Does this create a renumbering when extra classes are added?

if so, this a trade-off between binding verbosity, and diff size on changes. I'm not sure which one is worse... 😄

P.S. For these kind of changes that are trade-offs, we should file issues and discuss if we want that. (Preferably _before_ the coding, so it saves coding time for you 🙃)

mahesh-hegde:
> Does this create a renumbering when extra classes are added?

Renumbering should be rare, because it only happens when class names collide. Otherwise just simple name is used.

In practice, java libraries do good job of ensuring unique class names, since they are also imported unqualified.

dcharkes:
> > Does this create a renumbering when extra classes are added?
> 
> Renumbering should be rare, because it only happens when class names collide. Otherwise just simple name is used.
> 
> In practice, java libraries do good job of ensuring unique class names, since they are also imported unqualified.

Okay, you have convinced me! 🙃

mahesh-hegde:
Since multiple files and _package.dart are better implemented in separate PRs, what are the major changes I have to do in this PR (as of now)?

1. Remove `un_nest`?. I am still on fence with this, because sooner or later some user will request this, and it's not much code.

In java long import names are never a problem because IDE auto imports and folds them. I don't know how mature dart tooling is for this. In VS code I get auto import, but not sure about other LSP clients. 

Also CLI users may be irritated with long paths, whereas in Java I hardly touch terminal except from project root.

Sorry for being stubborn on this.

dcharkes:
I prefer only adding things if we really have a use for it, otherwise we just bloat our codebase. Especially while where still in the process of figuring some other aspects out.

So far I've not found it troublesome. VSCode just deals with this.

image


> Since multiple files and _package.dart are better implemented in separate PRs, what are the major changes I have to do in this PR (as of now)?
> 
> 1. Remove `un_nest`?

Yes, lets remove it and merge it. 👌

mahesh-hegde:
Removed `root_package`.

dcharkes (approved):

dcharkes:
🚀 🌔

PR #93, opened 2022-10-08
Move output related configuration to single place

### LINKED ISSUE #88: Output config structure
> * Added `root_package` option, which allows simpler hierarchy if all classes are from same package. For example, instead of `org/apache/pdfbox/pdmodel.dart` we can create `pdmodel.dart`, by telling that root_package is `org.apache.pdfbox`.

I presume we add this because it is annoying for people. 😄 
Though, Java explicitly doesn't support this.
The question is should we make our bindings look mostly like Java or mostly like Dart? (With ffigen in C, we say that when you write code against generated bindings, you're writing C in Dart syntax. So we're okay with violating dart naming for identifiers etc. With that reasoning, maybe we should stick with the nested folders, because it is more Java-esque.)

I would like to unify everything that influences _only_ output structure in the config:

```yaml
output:
  dart:
    path: 'lib/src/third_party/pdfbox/'
    output_structure: single_file || package_structure
  c:
    ## Root for generated C bindings.
    path: 'src/'
```

Also, maybe we could clean up the C output

```yaml
output:
  dart:
    path: 'lib/src/third_party/pdfbox/'
    output_structure: single_file || package_structure
  c:
    path: 'src/third_party/' # both generated C file and makefile?
    cmake_path: 'src/' # To make Flutter happy?
```

_Originally posted by @dcharkes in https://github.com/dart-lang/jnigen/issues/87#issuecomment-1268245201_

mahesh-hegde:
I think this closes #88


.github/workflows/test-package.yml
```diff
@@ -331,7 +331,7 @@ jobs:
       - run: flutter build apk
         working-directory: ./jnigen/example/notification_plugin/example
       - name: re-generate bindings
-        run: flutter pub run jnigen -Ddart_root=_dart -Dc_root=_c --config jnigen.yaml
+        run: flutter pub run jnigen -Doutput.dart.root=_dart -Doutput.c.root=_c --config jnigen.yaml
```
dcharkes, line 334: Should we call it `path` maybe instead of `root`? A path is necessarily a root if we output more than one file. And a path makes more sense if we output only a single file.

dcharkes (approved):

dcharkes:
Looks good to me! 🚀

.github/workflows/test-package.yml
mahesh-hegde: Should this be changed in the PR adding single file support or in current PR? Currently root is unambiguous and communicates it's a directory.

dcharkes: We're changing the API already in this PR. So if we do it now, we don't have to change the API twice.

Issue #91, opened 2022-10-07
jni:setup should generate bindings for all native dependencies.


mahesh-hegde:
Now `jni:setup` builds `package:jni`'s DLL, since the other packages are structured the same way w.r.t native bindings, we are using same script with `-p` flag to specify another package.

`dart run jni:setup && dart run jni:setup -p pdfbox_plugin`

However, this is not neat.

I think by default, `setup` should go through dependency list, and build the packages with `ffiPlugin` specified in pubspec. We can always provide it custom folders or package names through CLI, but default should be imitating what flutter's bundling does.

Also, should this script be in `jnigen` instead of `jni`?

dcharkes:
> I think by default, `setup` should go through dependency list, and build the packages with `ffiPlugin` specified in pubspec. We can always provide it custom folders or package names through CLI, but default should be imitating what flutter's bundling does.

I think relying on `ffiPlugin` is fragile because it doesn't enable Dart standalone. Also, it could be something else than jnigen generated code when it is an `ffiPlugin`.
`src/MakeFile.txt` is also fragile, because it could be something else than jnigen. And jnigen could be configured to use a different path than `src/MakeFile.txt`

That being said, having something automatic would be nice indeed.

How about we add a "Generated by package:jnigen." comment to the `src/CMakeLists.txt` and check both for that file _and_ a comment in that file? That way we don't accidentally build something that we didn't generate.

If devs output to a different path, I don't really know how to solve that, yet.

> Also, should this script be in `jnigen` instead of `jni`?

I think that makes sense indeed. jnigen is our main name. jni is the helper lib.

dcharkes:
> If devs output to a different path, I don't really know how to solve that, yet.

I have been working on standardizing `bin/setup.dart` or `bin/native.dart` as a CLI for native builds in general. And then `dart run` and `dart compile` (and even `flutter build`) would invoke that standardized CLI for all dependencies that have it. When that infrastructure is ready, we could make jnigen generate a `bin/native.dart` that has a single line invocation of the `src/CMakeLists.txt` (when can then actually generate the right path and find it if it was in a different output path), and the jnigen:setup step wouldn't have to do the packagesConfig traversal anymore because `dart` and `flutter` would do that.

For now, because I haven't standardized this CLI yet, I think your proposed solution is a good one and we should land it.

mahesh-hegde:
Few questions before I implement it.

1. Comment format

[...]

PR #96, opened 2022-10-13
Make jni:setup build all native libraries

mahesh-hegde:
Motivation explained in #91 .

This code is checking for a src/CMakeLists.txt in every package. Should it check for ffiPlugin key in pubspec instead?

### DIFF
--- a/jni/bin/setup.dart
+++ b/jni/bin/setup.dart
@@ -10,6 +10,8 @@ import 'package:package_config/package_config.dart';
 const ansiRed = '\x1b[31m';
 const ansiDefault = '\x1b[39;49m';
 
+const jniNativeBuildDirective = '# jnigen_native_build';
+
 const _defaultRelativeBuildPath = "build/jni_libs";
 
 const _buildPath = "build-path";
@@ -103,8 +105,6 @@ Future<String> findSources(String packageName) async {
 
 /// Return '/src' directories of all dependencies which has a CMakeLists.txt
 /// file.
-///
-/// TODO(PR): Should this rather check for ffiPlugin property of pubspecs?
 Future<Map<String, String>> findDependencySources() async {
   final packageConfig = await findPackageConfig(Directory.current);
   if (packageConfig == null) {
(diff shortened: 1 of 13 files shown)


dcharkes (commented):
I left comments on https://github.com/dart-lang/jnigen/issues/91.

jni/src/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+# jnigen_native_build
```
dcharkes, line 1: Maybe this should be

```
# Generated by `package:jnigen`.
# Special string to have it built by `dart run jni:setup`: "jnigen_native_build".
```

This way it's obvious why it is there, and users will not delete it.

dcharkes (approved):

jni/src/CMakeLists.txt
mahesh-hegde: `package:jni` itself is not generated.

Can we shorten it so that it's obvious while not too sentence-y.

something similar to `// ignore_for_file:` directives

mahesh-hegde: How about `# jni_native_build (Build with jni:setup. Do not delete this line).

dcharkes: sgtm

mahesh-hegde:
Done

jni/bin/setup.dart
```diff
@@ -10,7 +10,8 @@ import 'package:package_config/package_config.dart';
 const ansiRed = '\x1b[31m';
 const ansiDefault = '\x1b[39;49m';
 
-const jniNativeBuildDirective = '# jnigen_native_build';
+const jniNativeBuildDirective =
+    '# jni_native_build (Build with jni:setup. Do not delete this line)';
```
dcharkes, line 14: Period at the end? 🙃

mahesh-hegde: My bad 🙈

mahesh-hegde: Done

PR #97, opened 2022-10-13
bump jackson-databind to 2.14.0-rc2

mahesh-hegde:


mahesh-hegde:
Seems like 2.14.0-rc1 adds a new provided dependency which is not downloaded properly using maven. It can be handled by specifying all transitive deps. But if dependabout alerts are gone, I'd prefer to leave it at 2.13.4 in tests.

dcharkes:
> Seems like 2.14.0-rc1 adds a new provided dependency which is not downloaded properly using maven. It can be handled by specifying all transitive deps.

That is annoying. 😒 

> But if dependabout alerts are gone, I'd prefer to leave it at 2.13.4 in tests.

I don't know how to see if dependabot alerts are gone before merging. I don't think dependabot distinguishes between test and non-test dependencies. Maybe @jonasfj knows.

mahesh-hegde:
For me, two alerts both point to the POM in summarizer source.

dcharkes:
We can merge this one and find out 🙊

mahesh-hegde:
I am 99% sure this is sufficient. Both alerts are on databind depended upon in jnigen/java/pom.xml

dcharkes (approved):

mahesh-hegde:
Are dependabout alerts gone?

PR #102, opened 2022-10-18
Reduce concatenation in binding generator code

mahesh-hegde:
Attempt to straighten some of the generator code by moving into template style and reducing conditional control flow in binding writing code.

1) It should be easier to figure out where a piece of code came from. Since most wrapper symbols are written using 1 or 2 StringBuffer.write() calls.
2) It should be easier to change the binding code.

### DIFF
--- a/jnigen/lib/src/bindings/c_bindings.dart
+++ b/jnigen/lib/src/bindings/c_bindings.dart
@@ -95,17 +95,18 @@ class CBindingGenerator {
     final envMethod =
         isACtor ? 'NewObject' : 'Call$ifStaticCall${callType}Method';
     final returnResultIfAny = _result(m);
-    s.write('jmethodID $methodID = NULL;\n'
-        'FFI_PLUGIN_EXPORT\n'
-        '$jniResultType $cMethodName($cMethodParams) {\n'
-        '    $_loadEnvCall'
-        '    ${_loadClassCall(classRef, getInternalName(c.binaryName))}'
-        '    load_${ifStaticMethodID}method($classRef, '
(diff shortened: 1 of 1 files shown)


jnigen/lib/src/bindings/c_bindings.dart
```diff
@@ -72,50 +72,51 @@ class CBindingGenerator {
...
+      ifAssignResult = '${getCType(javaReturnType)} _result = ';
     }
-    s.write(_result(m));
-    s.write('}\n');
+
+    final ifStaticCall = isStatic ? 'Static' : '';
+    final envMethod =
+        isACtor ? 'NewObject' : 'Call$ifStaticCall${callType}Method';
+    final returnResultIfAny = _result(m);
+    s.write('jmethodID $methodID = NULL;\n'
```
liamappelbe, line 98: This would be even clearer as a multi-line string:

```dart
s.write('''
jmethodID $methodID = NULL;
FFI_PLUGIN_EXPORT
$jniResultType $cMethodName($cMethodParams) {

...

  ''');
```

mahesh-hegde: One thing I find annoying with multiline strings is we have to indent back. Is there  a rule of thumb when to use multiline strings and when to use normal strings?

liamappelbe: The rule of thumb is probably just "which approach makes my code clearer to read?"

liamappelbe (approved):

dcharkes:
Thanks @mahesh-hegde and @liamappelbe !

PR #108, opened 2022-10-26
fix: pr_checks now works on macOS

### LINKED ISSUE #41: Mac OS support
I don't have a Mac OS setup. Therefore unable to build dartjni for macos on desktop.

It's build system is different (podspec), and it may take more effort than porting same CMake config to windows.

I searched through github and sourcegraph, and unfortunately not able to find a single podspec that makes use of JNI libraries.

So this would involve finding JNI libraries and listing them in podspec then fixing os-specific errors / warnings..

HosseinYousefi:
Addressing parts of #41.

### DIFF
--- a/jnigen/tool/pr_checks.dart
+++ b/jnigen/tool/pr_checks.dart
@@ -16,26 +16,47 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:args/args.dart';
+
 import 'command_runner.dart';
 
-void main() async {
+// Flags
+const _clone = 'clone';
+
+void main(List<String> arguments) async {
+  final parser = ArgParser()
+    ..addFlag(
+      _clone,
+      help:
+          'Git clone will be run and the checks will happen in a temporary directory.',
+      defaultsTo: true,
+      negatable: true,
+    );
+  final argResults = parser.parse(arguments);
+  final shouldClone = argResults[_clone] as bool;
   final gitRoot = getRepositoryRoot();
 
   // change to project root
   Directory.current = gitRoot.toFilePath();
 
-  final tempDir = Directory.current.createTempSync('jnigen_checks_clone_');
-  final tempJniPath = tempDir.uri.resolve("jni/");
-  final tempJnigenPath = tempDir.uri.resolve("jnigen/");
-  final gitClone = Runner("Clone jni", Directory.current.uri)
-    ..chainCommand('git', ['clone', '.', tempDir.path])
-    ..chainCommand("flutter", ["pub", "get", "--offline"],
-        workingDirectory: tempJniPath)
-    ..chainCommand("dart", ["pub", "get", "--offline"],
-        workingDirectory: tempJnigenPath);
-  await gitClone.run();
+  var tempDir = Directory.current;
+  if (shouldClone) {
+    tempDir = Directory.current.createTempSync('jnigen_checks_clone_');
+    final gitClone = Runner("Clone jni", Directory.current.uri)
+      ..chainCommand('git', ['clone', '.', tempDir.path]);
+    await gitClone.run();
+  }
 
-  final jniAnalyze = Runner("Analyze JNI", tempJniPath);
+  final jniPath = tempDir.uri.resolve("jni/");
+  final jnigenPath = tempDir.uri.resolve("jnigen/");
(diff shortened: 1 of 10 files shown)


dcharkes (changes requested):
Please add MacOS to .github/workflows/test-package.yml

See https://github.com/dart-lang/ffigen/blob/master/.github/workflows/test-package.yml for inspiration on how to setup Mac bots.

jnigen/tool/pr_checks.dart
```diff
@@ -16,26 +16,47 @@
...
-void main() async {
+// Flags
+const _clone = 'clone';
+
+void main(List<String> arguments) async {
+  final parser = ArgParser()
+    ..addFlag(
+      _clone,
+      help:
+          'Git clone will be run and the checks will happen in a temporary directory.',
```
mahesh-hegde, line 31: Nit: I think it's conventional to write an imperative statement here.

`Run checks in a cloned copy of the project.`

jnigen/tool/pr_checks.dart
```diff
@@ -60,19 +81,18 @@ void main() async {
...
-        ..chainCommand("dart", ["run", "ffigen", "--config", "ffigen.yaml"])
-        ..chainCallback("compare bindings", () async {
-          final newBindingsText = await ffigenBindings.readAsString();
-          if (newBindingsText != oldBindingsText) {
-            await ffigenBindings.writeAsString(oldBindingsText);
-            throw "new JNI.h bindings differ from old bindings";
-          }
-        });
+  final ffigenCompare = Runner("Generate & Compare FFIGEN bindings", jniPath)
+    ..chainCommand("dart", ["run", "ffigen", "--config", "ffigen.yaml"])
```
mahesh-hegde, line 85: I think a small issue is that if the generated bindings are different, old ones are not restored.

If this is problematic, I think adding a `finally` block which writes back the old bindings may solve the issue.

mahesh-hegde (approved):

mahesh-hegde:
First of all, awesome work @HosseinYousefi.

If I understand correctly jni:setup works on dart standalone.

But since flutter FFIPlugin mechanism is different, we might have to do more things in order to get a flutter desktop using JNI running on macos.

However, it's not that important, because main use on desktop will always be during development use.

Re: checks, you may want to

1. Add a check that runs jni and jnigen tests on macos.

or

refactor existing checks for `jni` and `jnigen` into matrix form. This will require some conditionals because of windows / macos specific things, but make the YAML a little better.

jnigen/tool/pr_checks.dart
HosseinYousefi: Much clearer than my statement! Will change it.

jnigen/tool/pr_checks.dart
HosseinYousefi: I'll add it for completeness.

jnigen/tool/pr_checks.dart
```diff
@@ -60,19 +81,18 @@ void main() async {
...
+    ..chainCallback("compare bindings", () async {
+      final newBindingsText = await ffigenBindings.readAsString();
+      if (newBindingsText != oldBindingsText) {
+        await ffigenBindings.writeAsString(oldBindingsText);
```
HosseinYousefi, line 89: @mahesh-hegde  You're actually already doing it here!

dcharkes:
> Closes https://github.com/dart-lang/jnigen/issues/41.

To call MacOS supported we need a list of things:

* MacOS CLI apps using some jnigen'ed code working
* MacOS Flutter desktop app using some jnigen'ed code working
* MacOS as host for running jnigen on some Java sources
* MacOS as host for Android Flutter app working

Currently, this PR only checks the `package:jni` tests on CI. Which is fine for this PR, but we should keep #41 open until we support it as much as Linux/Windows.

jnigen/tool/pr_checks.dart
```diff
@@ -16,26 +16,46 @@
 import 'dart:async';
```
dcharkes, line 16: Should we run this script on the CI to check that we don't break the script accidentally?

Or does this try do things not supported on github bots?

dcharkes (approved):

dcharkes:
Thanks @HosseinYousefi ! 🎸

PR #110, opened 2022-10-26
Pure dart poc

mahesh-hegde:
This is PoC for pure dart bindings. 

There may be mistakes and fundamental oversights so take time to review it.

This changes jackson_core_test to use `dart_only` bindings type. Other tests and examples remain C based.


jnigen/lib/src/bindings/pure_dart_bindings.dart
```diff
@@ -0,0 +1,259 @@
...
+
+    writeAccessor(setter: false);
+    if (!isFinalField(f)) writeAccessor(setter: true);
+    return s.toString();
+  }
+
+  static const autoGeneratedNotice = '// Autogenerated by jnigen. '
+      'DO NOT EDIT!\n\n';
+  static const defaultLintSuppressions =
+      '// ignore_for_file: camel_case_types\n'
```
dcharkes, line 220: sort ascending

dcharkes (commented):
Nice! I also like that we have both options now, so that we could battle-test and benchmark them both.

It doesn't generate C but we still rely on `dartjni.c` being compiled, correct? To do the AttachThread etc.

> fundamental oversights

Nothing comes to mind. And if the tests using the bindings succeed, then we're not in a "computer says no" situation. 🙃

mahesh-hegde:
> It doesn't generate C but we still rely on dartjni.c being compiled, correct? To do the AttachThread etc.

Yes. Method call and field access infrastructure is provided by package:jni only.

The main advantage is you don't have to compile entire C code.With large C libraries size of DLL is also a concern but (at least in theory) dart bindings can be tree shaked.

Performance will not be good if you compare until variadics land and `PureDartBindingsGenerator` is adapted appropriately. Because we are using `Call<type>MethodA` which needs an array argument, which is allocated, and arguments filled using `jvalues` which checks runtimeType of each argument. I don't know a better way here.

dcharkes (approved):
sgtm

We need documentation in the readme for the new key.
Please list the trade-offs there as well.

mahesh-hegde:
We don't have a coherent documentation for YAML keys.

Should I address that in separate PR or current one?

dcharkes:
> We don't have a coherent documentation for YAML keys.
> 
> Should I address that in separate PR or current one?

sgtm

You can steal ideas from the ffigen repo. That one is pretty good. 😄

PR #116, opened 2022-10-28
Summarizer fix

mahesh-hegde:
Fix doclet summarizer generating private and package private methods.


jnigen/java/src/test/resources/com/example/Example.java
```diff
@@ -5,28 +5,28 @@
...
   }
 
   public static class Aux extends Example {
-    static int nothing = 0;
+    public static int nothing = 0;
 
-    static Example getAnExample() {
+    public static Example getAnExample() {
       return new Example();
     }
```
dcharkes, line 31: Do we need to add a `private`, `protected`, and no-modifier method/field as well?

Also, what can and can't we call through JNI? Only `public`?

dcharkes (approved):

jnigen/java/src/test/resources/com/example/Example.java
mahesh-hegde: It looks to be same as reflection, in that you can call non public methods. But we wouldn't want to expose private APIs either way.

dcharkes: Well ... maybe users do want to do that at some point.

Shouldn't we just report all in the summarizer? (E.g. the summarizer should not be opinionated, it should just summarize.)

And then we filter on public in jnigen. And then if some user comes around that wants the private stuff, we add a flag.

mahesh-hegde: > Shouldn't we just report all in the summarizer?

Sounds like technically correct idea.

What should I do now? Implement it in current PR or add to issues?

mahesh-hegde: I have pushed a commit.

dcharkes (approved):

dcharkes:
Squashing one bug at the time! 🐛

PR #118, opened 2022-11-03
Array support

### LINKED ISSUE #762: Typed array accessors
Currently accessible through `Jni.env` methods, however we should create a better API.

Likely, we would want an API similar to:

* https://api.dart.dev/dev/2.19.0-240.0.dev/dart-ffi/Pointer-class.html
* https://api.dart.dev/dev/2.19.0-240.0.dev/dart-ffi/Array-class.html

API features:

* `[]` and `[]=`
* Nesting (e.g. `int[][][]`)

HosseinYousefi:
Array support for primitives and objects. This is just the first prototype, feel free to challenge things fundamentally!


jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,212 @@
...
+    expect(array[2], 7);
+    expect(() {
+      array[-1] = 4;
+    }, throwsRangeError);
+    expect(() {
+      array[3] = 4;
+    }, throwsRangeError);
+    array.delete();
+  });
+  test("Java short array", () {
```
mahesh-hegde, line 78: There seems to be some redundancy here. Can this somehow be extracted into an interface so that the common code can be factored into a function?

jnigen/lib/src/bindings/common.dart
```diff
@@ -141,7 +159,7 @@ abstract class BindingsGenerator {
     const primitives = {
       'byte': 'Int8',
       'short': 'Int16',
-      'char': 'Int16',
+      'char': 'Uint16',
```
mahesh-hegde, line 162: Good catch.

mahesh-hegde (commented):
A cursory review. I will make another pass tomorrow.

Overall, great work!

jni/lib/src/jni_array.dart
```diff
@@ -0,0 +1,292 @@
...
+    return _length ??= _env.GetArrayLength(reference);
+  }
+}
+
+extension NativeJniArray<E extends NativeType> on JniArray<E> {
+  void _allocate(int size, void Function(Pointer<E> ptr) use) {
+    using((arena) {
+      final ptr = arena.allocate<E>(size);
+      use(ptr);
+    }, malloc);
```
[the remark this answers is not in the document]
HosseinYousefi: Isn't `calloc` just `malloc` that makes everything `0`. There is no need for that in this case as we're going to fill it up ourselves.

jni/lib/src/jni.dart
```diff
@@ -190,6 +190,75 @@ abstract class Jni {
...
+  /// Constructs a double array with the given length.
+  static JniArray<JDouble> newDoubleArray(int length) {
+    final ref =
+        accessors.newPrimitiveArray(length, JniType.doubleType).checkedRef;
+    return JniArray.fromRef(ref);
+  }
+
+  /// Constructs an Object array with the given length.
+  /// All elements are initially set to the element at initialElementRef.
+  static JniArray<T> newArray<T extends JniObject>(
```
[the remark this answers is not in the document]
HosseinYousefi: Perhaps we change the API to have a single `newArray` for both objects and primitives, if not I agree with `newObjectArray` for consistency.

jni/test/jni_array_test.dart
HosseinYousefi: For now it helped me identify a bug I had. We have an extension for each primitive type, it's difficult to make a generic common code that uses `[]` and `[]=` for the same reason that we couldn't have it without multiple extensions.

jnigen/example/notification_plugin/lib/notifications.dart
```diff
@@ -51,3 +53,14 @@ class Notifications extends jni.JniObject {
               text.reference)
           .check();
 }
+
+extension NotificationsJniArray on jni.JniArray<Notifications> {
```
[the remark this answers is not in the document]
HosseinYousefi: You may still want to create an array of objects for the sake of it inside Dart even if the array is never used anywhere in Java.

jnigen/lib/src/bindings/pure_dart_bindings.dart
```diff
@@ -218,24 +219,27 @@ class PureDartBindingsGenerator extends BindingsGenerator {
...
       '// ignore_for_file: unused_import\n'
       '// ignore_for_file: unused_shown_name\n'
       '\n';
 
+  static const ffiImport = 'import "dart:ffi" as ffi;\n';
   static const jniImport = 'import "package:jni/jni.dart" as jni;\n\n';
   static const internalHelpersImport =
       'import "package:jni/internal_helpers_for_jnigen.dart";\n\n';
-  static const defaultImports = jniImport + internalHelpersImport;
+  static const defaultImports = ffiImport + jniImport + internalHelpersImport;
```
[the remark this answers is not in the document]
HosseinYousefi: It is only used for `JniArray<primitives>`. I could generate `jni.JniArray<jni.JInt>` instead of `jni.JniArray<ffi.Int32>` and we wouldn't need to import ffi.

jni/src/dartjni.c
```diff
@@ -368,6 +368,122 @@ JniResult newObject(jclass cls, jmethodID ctor, jvalue* args) {
...
+      value.c = ((*jniEnv)->GetCharArrayElements(jniEnv, array, NULL))[index];
+      break;
+    case intType:
+      value.i = ((*jniEnv)->GetIntArrayElements(jniEnv, array, NULL))[index];
+      break;
+    case longType:
+      value.j = ((*jniEnv)->GetLongArrayElements(jniEnv, array, NULL))[index];
+      break;
+    case floatType:
+      value.f = ((*jniEnv)->GetFloatArrayElements(jniEnv, array, NULL))[index];
```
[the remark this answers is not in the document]
mahesh-hegde: Also I think it's better to 
1) get array regions instead of entire array elements

Or

2) Use `{Get,Release}PrimitiveArrayCritical` because we aren't blocking in between Get and Release.

It seems we cant get a `region` in critical section, unfortunately.

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,212 @@
...
+    expect(array[2], false);
+    array.setRange(0, 3, [false, true, true, true], 1);
+    expect(array[0], true);
+    expect(array[1], true);
+    expect(array[2], true);
+    expect(() {
+      array[-1] = false;
+    }, throwsRangeError);
+    expect(() {
+      array[3] = false;
```
mahesh-hegde, line 32: `_ = array[3]` if you want to ensure both getter and setter are checking range errors.

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,212 @@
...
+  test("Java boolean array", () {
+    final array = Jni.newBooleanArray(3);
+    expect(array.length, 3);
+    array[0] = true;
+    array[1] = false;
+    array[2] = false;
+    expect(array[0], true);
+    expect(array[1], false);
+    expect(array[2], false);
+    array.setRange(0, 3, [false, true, true, true], 1);
```
mahesh-hegde, line 24: Tangential: I wonder we can interop better with `typed_data` here.

jni/test/jni_array_test.dart
mahesh-hegde: Yeah it seems extensions can't implement interfaces nevermind.

jnigen/example/in_app_java/lib/android_utils.dart
```diff
@@ -36,6 +38,17 @@ class AndroidUtils extends jni.JniObject {
       _showToast(mainActivity.reference, text.reference, duration).check();
 }
 
+extension AndroidUtilsJniArray on jni.JniArray<AndroidUtils> {
+  AndroidUtils operator [](int index) {
+    return AndroidUtils.fromRef(
+        elementAt(index, jni.JniType.objectType).object);
```
mahesh-hegde, line 44: Maybe elementAt can call `object` getter? Are you explicitly checking `JniResult.exception` somewhere?

jnigen/example/in_app_java/lib/android_utils.dart
```diff
@@ -36,6 +38,17 @@ class AndroidUtils extends jni.JniObject {
...
+  }
+
+  void operator []=(int index, AndroidUtils value) {
+    (this as jni.JniArray<jni.JniObject>)[index] = value;
```
mahesh-hegde, line 48: I am still surprised this casting is allowed haha.

jnigen/lib/src/bindings/common.dart
```diff
@@ -92,6 +94,18 @@ abstract class BindingsGenerator {
...
   }
 
+  String dartArrayExtension(ClassDecl decl) {
+    final name = decl.finalName;
+    return '\nextension ${name}JniArray on ${jni}JniArray<$name> {\n'
+        '$indent$name operator [](int index) {\n'
+        '${indent * 2}return $name.fromRef(elementAt(index, ${jni}JniType.objectType).object);\n'
+        '$indent}\n\n'
+        '${indent}void operator []=(int index, $name value) {\n'
+        '${indent * 2}(this as ${jni}JniArray<${jni}JniObject>)[index] = value;\n'
```
mahesh-hegde, line 104: `this as $jniArrayType<$jniObjectType>`?

jnigen/test/bindings_test.dart
```diff
@@ -95,6 +101,22 @@ void main() async {
...
+    final ex2 = Example();
+    ex1.setInternal(1);
+    ex2.setInternal(2);
+    final array = Jni.newArray<Example>(2, ex1.getClass().reference);
+    array[0] = ex1;
+    array[1] = ex2;
+    print(array[0].getInternal());
+    expect(array[0].getInternal(), 1);
+    expect(array[1].getInternal(), 2);
+    array.delete();
```
mahesh-hegde, line 115: do we have some kind of `deleteAll`? or you can use a scoped Arena and do

```dart
final ex1 = Example()
    ..deletedIn(arena);
```

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
```diff
@@ -920,7 +923,7 @@ class JsonFactory extends jni.JniObject {
   /// Method for constructing parser for parsing
   /// the contents of given byte array.
   ///@since 2.1
-  jsonparser_.JsonParser createParser4(jni.JniObject data) =>
+  jsonparser_.JsonParser createParser4(jni.JniArray<ffi.Int8> data) =>
```
mahesh-hegde, line 926: @dcharkes Java Byte is signed, but it might make more sense to map `byte[]` to `jniArray<ffi.Uint8>`?

dcharkes (commented):
some initial comments

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
HosseinYousefi: One advantage that the current situation has is that every `JPrimitive` maps to a different FFI type, I'm using this fact to for example return a `bool` for an array of `JBoolean` (`Uint8`) and return an `int` for an array of `JByte` (`Int8`).

jni/pubspec.yaml
```diff
@@ -8,6 +8,7 @@ environment:
   flutter: ">=2.11.0"
 
 dependencies:
+  collection:
```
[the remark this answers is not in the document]
HosseinYousefi: I'll add a suitable one!

jnigen/example/in_app_java/lib/android_utils.dart
HosseinYousefi: I'm also using the same `elementAt` for primitives, I could have a `_primitiveElementAt` for those and another `elementAt(index)` that I expose for other classes to use. The problem with that is you could mistakenly write `primitiveArray.elementAt(3)` and get an error and not know why!

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,212 @@
...
+      array[3] = "44".jniString();
+    }, throwsRangeError);
+    array.delete();
+  });
+  test("Java 2d array", () {
+    final array = Jni.newIntArray(3);
+    array[0] = 1;
+    array[1] = 2;
+    array[2] = 3;
+    final twoDimArray = Jni.newArray<JniArray<JInt>>(
```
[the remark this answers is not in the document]
HosseinYousefi: Good idea, I'll try it.

jni/lib/src/accessors.dart
```diff
@@ -76,6 +76,11 @@ extension JniIdLookupResultMethods on JniPointerResult {
     _check(exception);
     return id.cast<jfieldID_>();
   }
+
+  Pointer<Void> get checkedRef {
```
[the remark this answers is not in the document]
HosseinYousefi: I'm returning `JniPointerResult` instead of `JniResult` though. Do you suggest changing it to `JniResult`?

jni/lib/src/jni.dart
```diff
@@ -190,6 +190,75 @@ abstract class Jni {
...
+      int length, Pointer<Void> classRef,
```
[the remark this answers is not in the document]
HosseinYousefi: It was `JniClass` in the previous commit haha

jni/lib/src/accessors.dart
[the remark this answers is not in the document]
mahesh-hegde: I have used JniPointerResult in case of ID lookups, if I recall correctly.

jni/lib/src/jni.dart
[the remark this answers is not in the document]
mahesh-hegde: Haha. That's one thing I learned the hard way. By keeping commits as small and atomic as possible, you get the ability to use `git revert`, `git bisect` etc..

jni/src/dartjni.c
```diff
@@ -368,6 +368,122 @@ JniResult newObject(jclass cls, jmethodID ctor, jvalue* args) {
...
+  attach_thread();
+  // Boundary checks
+  if (index < 0 || index >= (*jniEnv)->GetArrayLength(jniEnv, array)) {
+    JniClassLookupResult classRes =
+        getClass("java/lang/ArrayIndexOutOfBoundsException");
+    if (classRes.exception) {
+      result.exception = classRes.exception;
+      return result;
+    }
+    JniPointerResult ctorRes = getMethodID(classRes.classRef, "<init>", "()V");
```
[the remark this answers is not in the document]
HosseinYousefi: I didn't get that. Can you elaborate?

jni/test/jni_array_test.dart
HosseinYousefi: Will add this 👍

jnigen/lib/src/bindings/common.dart
HosseinYousefi: Cleaner 👍

jni/src/dartjni.c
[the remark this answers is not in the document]
mahesh-hegde: > Index 4 out of bounds for length 4

I don't see that information passed anywhere.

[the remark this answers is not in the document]
HosseinYousefi: I see. The problem is `Get<Primitive>ArrayElements` gives a pointer to the first element, so `ptr[index]` would SEGV instead of throwing an exception.

jni/src/dartjni.c
```diff
@@ -368,6 +368,122 @@ JniResult newObject(jclass cls, jmethodID ctor, jvalue* args) {
...
+}
+
+JniResult getArrayElement(jarray array, int index, int type) {
+  JniResult result = {NULL, NULL};
+  attach_thread();
+  // Boundary checks
+  if (index < 0 || index >= (*jniEnv)->GetArrayLength(jniEnv, array)) {
+    JniClassLookupResult classRes =
+        getClass("java/lang/ArrayIndexOutOfBoundsException");
+    if (classRes.exception) {
```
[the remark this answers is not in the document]
HosseinYousefi: Answered in the other comment.

jni/src/dartjni.c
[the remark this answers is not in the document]
HosseinYousefi: Changing this to use `ArrayRegion` instead might fix it and in that case I won't have to create my own exception.

jni/lib/src/jni_array.dart
```diff
@@ -0,0 +1,292 @@
...
+    });
+  }
+}
+
+extension ObjectJniArray<T extends JniObject> on JniArray<T> {
+  JniObject operator [](int index) {
+    return JniObject.fromRef(elementAt(index, JniType.objectType).object);
+  }
+
+  void operator []=(int index, JniObject value) {
```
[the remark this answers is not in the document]
HosseinYousefi: Unlike `elementAt` the logic of setting is not the same between primitives and objects yet (since I did not create an accessor for it).

We could even write`setRange(index, index + 1, [value])` without the need of any other function if you prefer.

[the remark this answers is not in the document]
HosseinYousefi: But that's going to introduce a tiny bit of overhead compared to what we currently have.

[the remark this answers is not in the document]
mahesh-hegde: More than tiny bit of overhead, there's allocation :(.

An idea might be have one accessor for objects, one for all integer types taking long, and use downcasting in C. That will reduce allocation burden in single set.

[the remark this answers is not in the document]
HosseinYousefi: Here's the `setRange` for Objects:

```dart
  void setRange(int start, int end, Iterable<JniObject> iterable,
      [int skipCount = 0]) {
    RangeError.checkValidRange(start, end, length);
    final size = end - start;
    final it = iterable.skip(skipCount).take(size);
    it.forEachIndexed((index, element) {
      this[index] = element;
    });
  }
  ```

It is simply calling `operator []=` for each element. Creating the extra list is the overhead, but that's it.

**Comment by HosseinYousefi** [comment]
I have made some changes, some other changes could be done:

1. `.arrayOf` getter to `JniTypeClass<T>` that returns `JniTypeClass<JniArray<T>>`
2. `.typeClass` static getter to each generated class that returns `XTypeClass`
   * We could then make `XTypeClass` itself private.
3. Currently name clashes could happen for classes like `X` and `XTypeClass`. If we decide to make type classes private like `_XTypeClass` this gets fixed.
4. `JniXTypeClass` is too long, but both `JniType` and `JniClass` were taken so I'm temporarily using it for consistency for the base class as well as the derived classes.
   * I propose using `JniType` and renaming `JniType` to something like `JniCallType`.

**Comment by HosseinYousefi** [comment]
Reminder to add const constructors to TypeClasses!

File: jni/test/jni_array_test.dart [hunk 22]
```diff
@@ -0,0 +1,232 @@
...
+    array[0] = '1';
+    array[1] = '2';
+    array[2] = '3';
+    expect(array[0], '1');
+    expect(array[1], '2');
+    expect(array[2], '3');
+    array.setRange(0, 3, ['4', '5', '6', '7'], 1);
+    expect(array[0], '5');
+    expect(array[1], '6');
+    expect(array[2], '7');
```
dcharkes, line 51: What about non-asci chars? This is int16.

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,232 @@
...
+    }, throwsRangeError);
+    array.delete();
+  });
+  test("Java float array", () {
+    final array = JniFloatTypeClass().newArray(3);
+    expect(array.length, 3);
+    array[0] = 0.5;
+    array[1] = 2;
+    array[2] = 3;
+    expect(array[0], 0.5);
```
dcharkes, line 141: Shouldn't these expect a roughly equals? e.g. floats/doubles are rarely bitwise equal

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,232 @@
...
+    expect(array[2].toDartString(), "33");
+    array.setRange(
+      0,
+      3,
+      ["44".jniString(), "55".jniString(), "66".jniString(), "77".jniString()],
+      1,
+    );
+    expect(array[0].toDartString(), "55");
+    expect(array[1].toDartString(), "66");
+    expect(array[2].toDartString(), "77");
```
dcharkes, line 200: Add tests with:

* Special chars
* Escaping

jni/test/jni_array_test.dart
dcharkes: Hehe yeah, but in general with some other numbers int might not.

(Especially for float because we convert double in Dart to a float in C which then goes to jni.)

I think it's just best practise to expect approx instead of precise, so it doesn't suddenly break on a slightly different setup/hardware.

HosseinYousefi: Sure I'll change it to approx!

mahesh-hegde:
Nice. Feel free to merge the later commits without waiting for my review.

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,237 @@
...
+  // Don't forget to initialize JNI.
+  if (!Platform.isAndroid) {
+    try {
+      Jni.spawn(dylibDir: "build/jni_libs", jvmOptions: ["-Xmx128m"]);
+    } on JvmExistsException catch (_) {
+      // TODO(#51): Support destroying and reinstantiating JVM.
+    }
+  }
+  test("Java boolean array", () {
+    final array = JniArray(const JniBooleanType(), 3);
```
dcharkes, line 20: JniArray -> JArray?
JniBooleanType -> JBooleanType? (and the others)

jni/test/jni_array_test.dart
```diff
@@ -0,0 +1,237 @@
...
+    }, throwsRangeError);
+    expect(() {
+      array[-1] = 4;
+    }, throwsRangeError);
+    expect(() {
+      array[3] = 4;
+    }, throwsRangeError);
+    array.delete();
+  });
+  const epsFloat = 1e-6;
```
dcharkes, line 139: nit: we usually write out names

https://dart.dev/guides/language/effective-dart/documentation#avoid-abbreviations-and-acronyms-unless-they-are-obvious

jni/ffigen.yaml
```diff
@@ -19,6 +19,9 @@ headers:
     - 'third_party/jni.h' # jni.h from Android NDK
 compiler-opts:
   - '-Ithird_party/'
+enums:
```
dcharkes, line 22: Just gotta love the package:ffigen features. 👌

jnigen/test/bindings_test.dart
```diff
@@ -95,6 +101,22 @@ void main() async {
     ex.delete();
   });
 
+  test('array of the class', () {
+    final ex1 = Example();
```
dcharkes, line 105: @HosseinYousefi an unnamed constructor 😮

dcharkes (approved):

jnigen/test/bindings_test.dart
HosseinYousefi: I know! Crazy! We had to create it ourselves 🥲

Issue #54, opened 2022-09-09
Helpers for constructing JNI signatures


dcharkes:
To construct strings such as `"I(I)"` we could add some helper code that could construct them from Dart function types.

```dart
  String typeToJavaSignature<T extends Function>() {
    final splitOnArrow = T.toString().split(' => ');
    final returnType = splitOnArrow[1];
    final arguments = splitOnArrow[0];
    final argumentsTrimmed = arguments.substring(1, arguments.length - 1);
    final argumentTypes = argumentsTrimmed.split(', ');
    final javaReturn = dartTypeToJavaSignatureType[returnType]!;
    final javaArguments =
        argumentTypes.map((e) => dartTypeToJavaSignatureType[e]!).join(',');
    return '($javaArguments)$javaReturn';
  }
```

```dart
final javaSignature = typeToJavaSignature<int Function(int)>();
```

The is probably not general enough, but maybe it would be a nice addition to the jni interface over having users having to construct the strings themselves.

dcharkes:
This probably cannot be fully shared with the logic in jnigen, because there we generate from the serialized Java type. But the logic in there should be similar to logic of doing it with Dart signatures.

We could consider jnigen generating Dart function types representing the signatures, that would make the API a bit more typed. Unfortunately we cannot change `T callStaticMethodByName<T>` to `T callStaticMethodByName<T, S extends T Function(...)>` constraining the return type and also not constraining the argument types which are `List<dynamic>`. However, we could runtime check the arguments based on the signature if we do `T callStaticMethodByName<T, S extends Function>`.

mahesh-hegde:
There are few practical difficulties in using function types. First, signatures use qualified names, which don't appear in string representation of pointer type. Second, one primitive type in dart may map to multiple types in java. (Int -> byte, short, int, long). There maybe more I am forgetting. So unless you're trying to get signature for simplest methods using int, float, string etc.. it won't work.

I also thought of inferring it from arguments passed. But that doesn't work because the actual type that function defines may be a superclass. But that may be a good default.

dcharkes:
> Second, one primitive type in dart may map to multiple types in java.

In `dart:ffi` we work around that by introducing type markers for C types `Int8 Function(Int8)`.

> signatures use qualified names

```dart
class Foo {}

const Map<Type, String> _fullyQualifiedNames = {
  Foo: "com.example.Foo",
};
```

[...]

PR #126, opened 2022-11-09
`JniX` to `JX` rename

### LINKED ISSUE #124: Define primitive types (`JBoolean` etc) manually
In https://github.com/dart-lang/jnigen/pull/118 creating an array of booleans is:

```dart
final array = JniArray(const JniBooleanType(), 3);
```

However, we would prefer

```dart
final array = JniArray(JBoolean.type, 3);
```

However, `JBoolean` is defined in in `jni.h` as a typedef, which is then translated to Dart:

```dart
typedef JBoolean = ffi.Uint8;
typedef JByte = ffi.Int8;
typedef JChar = ffi.Uint16;
typedef JShort = ffi.Int16;
typedef JInt = ffi.Int32;
typedef JLong = ffi.Int64;
typedef JFloat = ffi.Float;
typedef JDouble = ffi.Double;
```

Instead of leaking these typedefs to the user, we should rename them in our ffigen config, and instead define `JBoolean` (and the others) ourselves, so that we can define a `.type` static getter that returns `JniType<JBoolean>`.

TODO:

- [ ] Check if this causes problems somewhere else. E.g. when we start generating code for callbacks and we have to deal with function signatures, do we cause problems for ourselves because those signatures would have the typedefs instead of our own type hierarchy?
- [ ] Check if it would be useful to introduce a `JPrimitive` that these 8 types implement.

### LINKED ISSUE #125: Rename `JniX` to `JX`
Motivation is to write shorter code, also java types are called `jint`, `jobject`, ... in C, which makes it familiar for people who have some experience with JNI.

HosseinYousefi:
Closes #124. Closes #125.


jni/test/jarray_test.dart
```diff
@@ -207,19 +207,19 @@ void main() {
...
-    final array = JniArray(const JniIntType(), 3);
+    final array = JArray(JInt.type, 3);
     array[0] = 1;
     array[1] = 2;
     array[2] = 3;
-    final twoDimArray = JniArray(JniArray.type(const JniIntType()), 3);
+    final twoDimArray = JArray(JArray.type(JInt.type), 3);
     expect(twoDimArray.length, 3);
     twoDimArray[0] = array;
     twoDimArray[1] = array;
```
dcharkes, line 225: Please add tests with `JArray<JObject>`.

jni/test/jarray_test.dart
```diff
@@ -186,18 +186,18 @@ void main() {
     array.delete();
   });
   test("Java string array", () {
-    final array = JniArray(JniString.type, 3);
+    final array = JArray(JString.type, 3);
     expect(array.length, 3);
-    array[0] = "حس".jniString();
-    array[1] = "\$".jniString();
-    array[2] = "33".jniString();
+    array[0] = "حس".toJString();
```
dcharkes, line 191: Since all of these are references, they rely on the attached finalizer and GC to be cleaned up.

It would be cleaner if we would eagerly clean up themselves. Consider using an Arena. (Also the other tests.)

I presume users will start looking at our examples and tests for how to use it, so lets teach them to clean up after themselves.

```dart
  test("Java string array", () {
    using((arena) {
      final array = JArray(JString.type, 3)..deletedIn(arena);
      expect(array.length, 3);
      array[0] = "حس".toJString()..deletedIn(arena);
      array[1] = "\$".toJString()..deletedIn(arena);
      array[2] = "33".toJString()..deletedIn(arena);
      expect(array[0].toDartString(), "حس");
      expect(array[1].toDartString(), "\$");
      expect(array[2].toDartString(), "33");
      array.setRange(
        0,
        3,
        [
          "44".toJString(),
          "55".toJString()..deletedIn(arena),
          "66".toJString()..deletedIn(arena),
          "77".toJString()..deletedIn(arena),
        ],
        1,
      );
      expect(array[0].toDartString(), "55");
      expect(array[1].toDartString(), "66");
      expect(array[2].toDartString(), "77");
      expect(() {
        final _ = array[-1];
      }, throwsRangeError);
      expect(() {
        array[-1] = "44".toJString()..deletedIn(arena);
      }, throwsRangeError);
      expect(() {
        array[3] = "44".toJString()..deletedIn(arena);
      }, throwsRangeError);
    });
  });
```

jni/lib/src/primitives.dart
```diff
@@ -0,0 +1,119 @@
...
+
+  @override
+  int get _type => JniCallType.byteType;
+
+  @override
+  String get signature => "B";
+}
+
+abstract class JBoolean extends JPrimitive {
+  static const type = _JBooleanType();
```
dcharkes, line 24: Don't leak out private types:

`static const JType<JBoolean> type = _JBooleanType();`

dcharkes:
Also takes a step in the direction of https://github.com/dart-lang/jnigen/issues/54.

jni/lib/src/array.dart
```diff
@@ -4,17 +4,25 @@
 
 // ignore_for_file: unnecessary_cast
 
-part of 'jni_object.dart';
+part of 'types.dart';
```
dcharkes, line 7: Some of the files in this folder have a plural name, and some singular.

I don't see any style guide on names: https://dart.dev/guides/language/effective-dart/style#do-name-libraries-and-source-files-using-lowercase_with_underscores

The Java style is a single class per file and have the filename correspond.
I believe we do the same in Dart often if we have a single class in a file.
If we have more than one class in a file, then we name it after the most important class if there is one. If there is no most important one it probably has a name reflecting the group of classes.

jarray.dart
jexception.dart (with a JException class that all other exceptions extend)
jobject.dart
jprimitives.dart
jreference.dart
jstring.dart

(Feel free to do this in a separate PR though. If so, file an issue, and we can do it for the whole project at once.)

dcharkes (approved):
Beautiful! (okay I might be biased! 🙃)

jni/lib/src/array.dart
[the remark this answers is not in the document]
HosseinYousefi: We didn't use `part of`, I added it! It allows us to have private members between the classes and still be able to use them internally.

HosseinYousefi: What about `types.dart`? I can add `j`s to all the names

jni/test/jarray_test.dart
HosseinYousefi: I'll change it.

jni/lib/src/primitives.dart
HosseinYousefi: Good catch!

jni/lib/src/array.dart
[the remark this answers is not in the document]
mahesh-hegde: If such private symbols are less in number, you can also hide them while exporting from library.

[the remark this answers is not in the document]
HosseinYousefi: For example in `array.dart` I use `typeClass._type` and `_type` is defined in `JniType` which resides in `types.dart`, importing would make this not work. Also can I "hide" a single property?

`part of` is used in dart quite often, for example in `core.dart`  parts `list`, `map`, `string`, and ...

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonToken.dart
```diff
@@ -40,28 +40,28 @@ import "../../../../_init.dart" show jniEnv, jniAccessors;
 ///
 /// Enumeration for basic token types used for returning results
 /// of parsing JSON content.
-class JsonToken extends jni.JniObject {
+class JsonToken extends jni.JObject {
   static final _classRef =
       jniAccessors.getClassOf("com/fasterxml/jackson/core/JsonToken");
-  JsonToken.fromRef(jni.JObject ref) : super.fromRef(ref);
+  JsonToken.fromRef(jni.JObjectPtr ref) : super.fromRef(ref);
```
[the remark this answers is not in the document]
HosseinYousefi: That is certainly possible to do. I will create another issue for it though.

jni/lib/src/array.dart
HosseinYousefi: `types.dart` here acts like `dart:core`'s `core.dart`. For now I'll add a `j` prefix and have all to be a part of `types.dart` and if you have some other suggestion, please file an issue.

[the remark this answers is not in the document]
dcharkes: Hiding `_type` is a good argument. I don't like `part of`, but I'm convinced.

Where is package visibility when you need it (something Java has done more right than Dart? 🙈 )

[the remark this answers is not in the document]
HosseinYousefi: Why don't you like `part of`? 🤔

PR #130, opened 2022-11-09
Added `JArray.filled`

### LINKED ISSUE #129: Add `JArray.filled` constructor
The api for JNI already supports passing an initial element for object arrays, but currently there is no way to provide this through `JArray(SomeJType, length)`. I propose using a similar syntax to `List.filled(length, element)`

```dart
final array = JArray.filled(JString.type, 3, someJString);
```

HosseinYousefi:
Closes #129.


dcharkes (approved):
lgtm

Thanks @HosseinYousefi ! 🚀

PR #150, opened 2022-11-23
Fixed the problem with annotations containing enums

HosseinYousefi:


dcharkes (approved):
lgtm! Squashing one 🐛  at the time, or is it 🔨  a mole? 🙃 

nit: Please write commit message in present tense: Fix ...

Also, please add a bug with an example of the issue that is fixed by this PR, or ensure that on merge you describe the issue in the commit message. And possibly on this PR as well. (Maybe add the error message/stack trace to the PR/bug from before this PR.)

(I reverse engineered the issue from the changes to jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/annotations/JsonSerializable.java, but that's hard to find if we look at this a year from now.)

HosseinYousefi:
The gist of the problem for the future wanderer:

Annotations such as `@Keep` that take no argument did work, but specifically annotations that get an enum value as argument failed, a test is created to show this failure: jnigen/test/simple_package_test/java/com/github/dart_lang/jnigen/annotations/JsonSerializable.java

The reason for the failure is converting these enum values to JSON is not possible automagically since there is no known serializer for it, so for now we're `.toString`ing it and show it in the properties of the annotation. The whole annotation enum logic could be improved, but there has already been TODOs pointing to #23 for that in the code base.

Issue #157, opened 2022-12-06
Better structure for APISummarizer


HosseinYousefi:
* Currently we have two parsers, one for sources and one for bytecode. Using metadata, these two will almost be the same with the exception of docs which could be fetched in case they're available. 
* The JSON export is very `string`y. It would be better for classes to have unique ids and actually pass in the full AST for an exact representation for the codegen. The codegen, uses a map of `String` to `ClassDecl` to resolve the classes. We need to have the same "resolving" logic for type bounds, generic type variables, etc. Using string names is not the way to go for all of these use-cases.
* I find the way `TypeUsage` has been declared to be confusing.

TL;DR: We would want to pass a tree instead of JSON objects and lookup values from a map (even if the format is still JSON).

dcharkes:
I believe it was string-y because identifiers are uniquely prefixed.
If references are resolved to the actual definitions, it will be a graph, not a tree, which is harder to serialize.

I have no objections against making the serialized format better.

Side note: because the ApiSummarizer is a separate process (rather than using jnigen itself to directly use the java packages for parsing) enables us to have different licensing for this separate program. Licenses are contagious when "linking" things together through FFI, but not when "invoking" a separate program through `Process.run`. So we can't use jnigen on the Java libraries and interact directly with their output without serialization step in between. (Which is what we do in ffigen, and what I had hoped we could do.)

@mahesh-hegde any thoughts based on your experience writing the summarizer?

mahesh-hegde:
> The JSON export is very stringy. It would be better for classes to have unique ids and actually pass in the full AST for an exact representation for the codegen.

Initially we only had declared classes. So having string keys referring to other classes made sense.

Also, this representation makes no difference between class in current set of classes vs some external class (which may be provided through an import map for example..). In my design, resolver object is supposed to take care of that.

> TypeUsage

It is basically serialised form of `TypeMirror` in doclet API. It's to differentiate between __declaration__ of a type vs __usage__, latter can have type parameters etc..

I have tried to mirror doclet API as close as possible. So ClassDecl is more analogue to `DeclaredType` than `TypeMirror`. A declared type is always in scope, but a type parameter or wildcard is only stored in a type usage and it's meaning depends on scope.

Re: comment. It's a fairly minor implementation detail. It's still a inheritance hierarchy where `PrimitiveType` etc.. extend `TypeUsage`. but deserializing it from JSON needs a hint of which type to deserialize into.

[...]

PR #158, opened 2022-12-07
Use .jar metadata to generate generic types

### LINKED ISSUE #132: Make asm/jar frontend support java type signatures from .jar file metadata (e.g. for generics)
We should make the asm/jar frontend to `package:jnigen` support java type signatures (see [spec](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3.4)).

Ideally we would have tests that ensures whether we run `package:jnigen` on `android.jar` or `android-src-stubs.jar` it produces exactly the same code modulo doc comments and parameter names. (one could do that by running jnigen on jar and src, building kernel outlines and comparing the two).

HosseinYousefi:
Closes #132.


jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmClassSignatureVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+  public void visitFormalTypeParameter(String name) {
+    var typeParam = new TypeParam();
+    typeParam.name = name;
+    decl.typeParams.add(typeParam);
+  }
+
+  @Override
+  public SignatureVisitor visitClassBound() {
+    var typeUsage = new TypeUsage();
+    decl.typeParams.get(decl.typeParams.size() - 1).bounds.add(typeUsage);
```
dcharkes, line 31: Why is it always the last from `decl.typeParams` ?

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmClassSignatureVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+    return new AsmTypeUsageSignatureVisitor(typeUsage);
+  }
+
+  @Override
+  public SignatureVisitor visitInterfaceBound() {
+    var typeUsage = new TypeUsage();
+    decl.typeParams.get(decl.typeParams.size() - 1).bounds.add(typeUsage);
```
dcharkes, line 38: ditto

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmMethodSignatureVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+  public void visitFormalTypeParameter(String name) {
+    var typeParam = new TypeParam();
+    typeParam.name = name;
+    method.typeParams.add(typeParam);
+  }
+
+  @Override
+  public SignatureVisitor visitClassBound() {
+    var typeUsage = new TypeUsage();
+    method.typeParams.get(method.typeParams.size() - 1).bounds.add(typeUsage);
```
dcharkes, line 31: ditto

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmMethodSignatureVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+    return new AsmTypeUsageSignatureVisitor(typeUsage);
+  }
+
+  @Override
+  public SignatureVisitor visitInterfaceBound() {
+    var typeUsage = new TypeUsage();
+    method.typeParams.get(method.typeParams.size() - 1).bounds.add(typeUsage);
```
dcharkes, line 38: ditto

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmTypeUsageSignatureVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+        break;
+      case 'S':
+        name = "short";
+        break;
+      case 'V':
+        name = "void";
+        break;
+    }
+    typeUsage.shorthand = name;
+    typeUsage.type = new TypeUsage.PrimitiveType(name);
```
dcharkes, line 56: Nit: should we not re-create `PrimitiveType`s for the same `name` every time?

(Feel free to ignore if it needlessly complicates the code.)

jnigen/example/in_app_java/jnigen.yaml
```diff
@@ -14,3 +14,5 @@ source_path:
 classes:
   - 'com.example.in_app_java.AndroidUtils' # from source_path
   - 'android.os.Build' # from gradle's compile classpath
+  - 'java.util.HashMap' # from gradle's compile classpath
```
dcharkes, line 17: Should we add a separate test instead of adding it on to an example?

If we release the standard library bindings as a separate package, we would likely want to remove HashMap from here again for example purposes. We shouldn't lose test coverage when doing that.

Maybe we should cover tests in `test/` and only add things to `example/` when we really want to illustrate something specific. And then it should probably be accompanied by a README.md detailing what the example shows.

dcharkes (approved):
lgtm

Thanks @HosseinYousefi !

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/disasm/AsmTypeUsageSignatureVisitor.java
HosseinYousefi: I plan to completely change the structure to be more tree like https://github.com/dart-lang/jnigen/issues/157. These will all change then. Parsing that structure would require writing our own parser but it's worth it and we won't have to depend of JsonSerializable.

jnigen/example/in_app_java/jnigen.yaml
HosseinYousefi: Good idea to separate the example. Will do that.

HosseinYousefi:
The issue with the test is interesting. We have a class that is defined like:

```java
public class JsonFactoryBuilder extends TSFBuilder<JsonFactory, JsonFactoryBuilder>
```

So it extends a class which has itself as a type parameter. Although the descriptor for this class is `com/fasterxml/jackson/core/JsonFactoryBuilder`, the signature becomes `Lcom/fasterxml/jackson/core/TSFBuilder<Lcom/fasterxml/jackson/core/JsonFactory;Lcom/fasterxml/jackson/core/JsonFactoryBuilder;>;`. This edge cases causes the asm backend test to fail.

jnigen/test/jackson_core_test/generate.dart
```diff
@@ -56,13 +56,16 @@ Config getConfig(
...
     exclude: BindingExclusions(
-      fields: excludeAll<Field>([
-        ['com.fasterxml.jackson.core.JsonFactory', 'DEFAULT_QUOTE_CHAR'],
-        ['com.fasterxml.jackson.core.Base64Variant', 'PADDING_CHAR_NONE'],
-        ['com.fasterxml.jackson.core.base.ParserMinimalBase', 'CHAR_NULL'],
-        ['com.fasterxml.jackson.core.io.UTF32Reader', 'NC'],
-      ]),
-    ),
+        fields: excludeAll<Field>([
+          ['com.fasterxml.jackson.core.JsonFactory', 'DEFAULT_QUOTE_CHAR'],
```
dcharkes, line 60: Do we need a TODO for these?

HosseinYousefi: These were existing before. Maybe?!

Issue #163, opened 2022-12-19
Field 'finalName' has not been initialized.


Talsoake:
I am using the asm backend to generate bindings for a jar file.
Writing the bindings fails with this stacktrace:

> (jnigen) INFO: Creating dart init file ...
Unhandled exception:
LateInitializationError: Field 'finalName' has not been initialized.
#0      Method.finalName (package:jnigen/src/elements/elements.dart)
#1      CBindingGenerator._method (package:jnigen/src/bindings/c_bindings.dart:73:20)
#2      CBindingGenerator._class (package:jnigen/src/bindings/c_bindings.dart:50:15)
#3      CBindingGenerator.generateBinding (package:jnigen/src/bindings/c_bindings.dart:35:42)
#4      MappedListIterable.elementAt (dart:_internal/iterable.dart:413:31)
#5      ListIterator.moveNext (dart:_internal/iterable.dart:342:26)
#6      new _GrowableList._ofEfficientLengthIterable (dart:core-patch/growable_array.dart:189:27)
#7      new _GrowableList.of (dart:core-patch/growable_array.dart:150:28)
#8      new List.of (dart:core-patch/array_patch.dart:51:28)
#9      ListIterable.toList (dart:_internal/iterable.dart:213:44)
#10     writeCBindings (package:jnigen/src/writers/bindings_writer.dart:65:55)
asynchronous suspension
#11     FilesWriter.writeBindings (package:jnigen/src/writers/files_writer.dart:217:7)
asynchronous suspension
#12     generateJniBindings (package:jnigen/src/generate_bindings.dart:106:5)
asynchronous suspension
#13     main (file:///home/user/snap/flutter/common/flutter/.pub-cache/git/jnigen-9c9c58fbaa2ce8a4a36b15199a26b2388da3763a/jnigen/bin/jnigen.dart:16:3)
asynchronous suspension
pub finished with exit code 255>

Cant figure out whats causing the issue here.

HosseinYousefi:
I might have an idea of what is causing this error. I opened a pull request #164, you can check out that branch and see if your problems is solved.

Talsoake:
Thanks for the quick response and the fix. I will test this next week.

Talsoake:
Now I had the time to test this.
Works like intended.
Issue is solved.

PR #164, opened 2022-12-21
Excluding methods and fields of the excluded classes.

HosseinYousefi:
Maybe closes? #163.


dcharkes (approved):
Please add a unit test exercising the faulty behavior.

PR #197, opened 2023-03-08
Use `package:cli_config`

dcharkes:
This PR replaces the command-line defines and yaml parsing logic with `package:cli_config`.

Notable changes:

* All paths are automatically resolved to absolute `Uri`s by `package:cli_config`.

Changes not done:

* Required keys are not marked as required at the moment, it would changes the control flow of the config. We could do this in a follow up CL.

Pre-merge todos:

* [x] Publish version v0.1.0 of `package:cli_config` and depend on that in this PR.

## Original proposal

I have seen at least three locations in which we want to handle configs:

1. package:ffigen
1. package:jnigen
1. https://github.com/dart-lang/sdk/issues/50565

So I propose we add a `package:config` that will be opinionated about how to provide config via CLI defines, environment variables, and/or a config file (YAML/JSON). I've made a prototype for this package here:

* https://github.com/dcharkes/config

Some of the logic is taken from this package, thanks @mahesh-hegde !

This PR is a proof of concept to check that `package:config` would satisfy the needs of `package:jnigen`.

Do not merge yet, I would like to iterate on `package:config` before releasing an initial version.

Feedback welcome.

One inconsistency I found between package:ffigen and package:jnigen is whether yaml keys use underscores or dashes.

Some related issues:

* https://github.com/dart-lang/jnigen/issues/186
* https://github.com/dart-lang/ffigen/issues/520

### DIFF
--- a/jnigen/lib/src/config/yaml_reader.dart
+++ b/jnigen/lib/src/config/yaml_reader.dart
@@ -104,13 +104,8 @@ class YamlReader {
     return configResult?.map((e) => e.path).toList();
   }
 
-  String? getOneOf(String property, Set<String> values) {
-    final value = _config.optionalString(property);
-    if (value == null || values.contains(value)) {
-      return value;
-    }
-    throw ConfigException('expected one of $values for $property');
-  }
+  String? getOneOf(String property, Set<String> values) =>
+      _config.optionalString(property, validValues: values);
 
   Map<String, String>? getStringMap(String property) {
     final value = _config.valueOf<YamlMap?>(property);
--- a/jnigen/lib/src/config/yaml_reader.dart
+++ b/jnigen/lib/src/config/yaml_reader.dart
@@ -6,15 +6,17 @@ import 'dart:io';
 
 import 'package:args/args.dart';
 import 'package:yaml/yaml.dart';
+import 'package:cli_config/cli_config.dart' as cli_config;
 
 import 'config_exception.dart';
 
 /// YAML Reader which enables to override specific values from command line.
 class YamlReader {
-  YamlReader.of(this.cli, this.yaml, this.yamlFile);
-  Map<String, String> cli;
-  Map<dynamic, dynamic> yaml;
-  File? yamlFile;
+  final cli_config.Config _config;
+
+  final Uri? _configRoot;
+
+  YamlReader.of(this._config, this._configRoot);
 
   /// Parses the provided command line arguments and returns a [YamlReader].
   ///
@@ -42,19 +44,15 @@ class YamlReader {
       stderr.writeln(parser.usage);
       exit(1);
     }
-    final configFile = results['config'] as String?;
-    Map<dynamic, dynamic> yamlMap = {};
-    if (configFile != null) {
+    final configFilePath = results['config'] as String?;
+    String? configFileContents;
+    Uri? configFileUri;
+    if (configFilePath != null) {
       try {
-        final yamlInput = loadYaml(File(configFile).readAsStringSync(),
-            sourceUrl: Uri.file(configFile));
-        if (yamlInput is Map) {
-          yamlMap = yamlInput;
-        } else {
-          throw ConfigException('YAML config must be set of key value pairs');
-        }
+        configFileContents = File(configFilePath).readAsStringSync();
+        configFileUri = File(configFilePath).uri;
       } on Exception catch (e) {
-        stderr.writeln('cannot read $configFile: $e');
+        stderr.writeln('Cannot read $configFilePath: $e.');
       }
     }
     final regex = RegExp('([a-z-_.]+)=(.+)');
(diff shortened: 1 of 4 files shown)


HosseinYousefi:
Good initiative @dcharkes! This can help users have a familiar experience when using our packages.

jnigen/lib/src/config/yaml_reader.dart
```diff
@@ -42,19 +44,15 @@ class YamlReader {
...
-        if (yamlInput is Map) {
-          yamlMap = yamlInput;
-        } else {
-          throw ConfigException('YAML config must be set of key value pairs');
-        }
+        configFileContents = File(configFilePath).readAsStringSync();
+        configFileUri = File(configFilePath).uri;
       } on Exception catch (e) {
-        stderr.writeln('cannot read $configFile: $e');
+        stderr.writeln('Cannot read $configFilePath: $e.');
```
mahesh-hegde, line 55: Shouldn't this be a fatal error?

jnigen/lib/src/config/yaml_reader.dart
```diff
@@ -69,97 +67,58 @@ class YamlReader {
...
+      _config.optionalPath(property)?.toFilePath();
 
-  List<String>? getStringList(String property) {
-    final configValue = cli[property]?.split(';') ??
-        getYamlValue<YamlList>(property)?.cast<String>();
-    return configValue;
-  }
+  List<String>? getStringList(String property) => _config.optionalStringList(
+        property,
+        splitCliPattern: ';',
```
mahesh-hegde, line 94: Here's room to some design. should we continue to have an ad-hoc format for string lists (separated by ';'), or use jsonDecode with another flag, allowing arbitrary JSON arguments?

jnigen/lib/src/config/yaml_reader.dart
```diff
@@ -69,97 +67,58 @@ class YamlReader {
...
-    final path = getYamlValue<String>(property);
-    if (path == null) return null;
-    // In (very unlikely) case YAML config didn't come from a file,
-    // do not try to resolve anything.
-    if (yamlFile == null) return path;
-    final yamlDir = yamlFile!.parent;
-    return yamlDir.uri.resolve(path).toFilePath();
-  }
+  String? getPath(String property) =>
+      _config.optionalPath(property)?.toFilePath();
```
mahesh-hegde, line 90: One choice here is, do command line overrides need to be resolved relative to current dir, or relative to YAML file? What does package:config do? In the old version CLI override is not resolved.

jnigen/lib/src/config/yaml_reader.dart
```diff
@@ -69,97 +67,58 @@ class YamlReader {
...
-    return paths.map((path) => yamlDir.uri.resolve(path).toFilePath()).toList();
+    final configResult = _config.optionalPathList(
+      property,
+      combineAllConfigs: false,
+      splitCliPattern: ';',
+    );
+    return configResult?.map((e) => e.path).toList();
   }
 
   String? getOneOf(String property, Set<String> values) {
```
mahesh-hegde, line 107: I think getOneOf can be moved to `package:config`, or a more structured one like `getEnum<T extends Enum>`.

jnigen/pubspec.yaml
```diff
@@ -2,13 +2,16 @@
...
+publish_to: none
+
 name: jnigen
 version: 0.3.0
 description: Experimental generator for FFI+JNI bindings.
 repository: https://github.com/dart-lang/jnigen/tree/main/jnigen
 
 environment:
-  sdk: '>=2.17.0 <4.0.0'
+  sdk: ">=2.17.0 <4.0.0"
```
mahesh-hegde, line 14: Ah we saw this before.

I set YAML formatting >> single quotes in VS code. Should not matter though.

mahesh-hegde (approved):
LGTM. My review is mostly questions and high level design choices.

jnigen/test/config_test.dart
```diff
@@ -95,8 +97,8 @@ void main() {
   final config = Config.parseArgs([
     '--config',
     jnigenYaml,
-    '-Doutput.c.path=$testSrc/',
-    '-Doutput.dart.path=$testLib/',
+    '-Doutput.c.path=$testSrc${Platform.pathSeparator}',
```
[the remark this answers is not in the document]
dcharkes: If we would have used `Uri` and `Directory` we would have had the path separator for free. String manipulation for paths just seems like a bad idea to me.

Anyway, I'm not here to argue about package:path vs Uri, but in my packages I use `Uri`. 😆

jnigen/test/config_test.dart
```diff
@@ -10,12 +12,12 @@ import 'package:path/path.dart' as path show equals;
 import 'jackson_core_test/generate.dart';
 
 const packageTests = 'test';
-final jacksonCoreTests = join(packageTests, 'jackson_core_test');
-final thirdParty = join(jacksonCoreTests, 'third_party');
-final lib = join(thirdParty, 'lib');
-final src = join(thirdParty, 'src');
-final testLib = join(thirdParty, 'test_', 'lib');
-final testSrc = join(thirdParty, 'test_', 'src');
+final jacksonCoreTests = absolute(packageTests, 'jackson_core_test');
```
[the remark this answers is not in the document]
dcharkes: The issue with relative is that it is unpredictable how the path will be used. So instead, `package:cli_config` always tries to resolve paths to absolute. (command-line defines are resolved relative to current working dir, paths in the config file are resolved relative to the config file absolute path.)

I feel log messages are less important than the behavior being predictable.

jnigen/lib/src/config/yaml_reader.dart
dcharkes: If you pass an override on the command line, you're likely auto-completing a path at your working directory, so resolve to current working dir.

jnigen/lib/src/config/yaml_reader.dart
dcharkes: I don't see a reason to break the current behavior. Unless there is something ugly about the current behavior.

jnigen/lib/src/config/yaml_reader.dart
dcharkes: Maybe, I tried to keep this PR behavior preserving.

mahesh-hegde: That should be an error anyway. If you make it an error, does any test or CI fail?

jnigen/lib/src/config/yaml_reader.dart
mahesh-hegde: Good point.

jnigen/lib/src/config/yaml_reader.dart
mahesh-hegde: It should be error anyway. My bad. Does any test/CI depend on this behavior?

jnigen/pubspec.yaml
```diff
@@ -2,13 +2,16 @@
 # for details. All rights reserved. Use of this source code is governed by a
 # BSD-style license that can be found in the LICENSE file.
 
+# DO NOT MERGE: prototype.
```
HosseinYousefi, line 5: No longer a prototype, right?

jnigen/pubspec.yaml
```diff
@@ -17,6 +20,11 @@ dependencies:
   args: ^2.3.0
   yaml: ^3.1.0
   logging: ^1.0.2
+  cli_config:
```
HosseinYousefi, line 23: ready to be 0.1.0

HosseinYousefi (approved):
LGTM! Thanks @dcharkes :)

PR #212, opened 2023-03-29
Generating JValue wrapper classes in pure Dart bindings

### LINKED ISSUE #211: Different result in pure Dart and Dart+C mode
The problem is with `toJValues` which calls `_fillJValue` and fills the `.i` (int) for Dart `int`. However Dart `int` is 64 bits so it should be converted to `.j`. We're not generating the correct `JValueLong` wrapper.

HosseinYousefi:
Closes #211.


dcharkes (approved):
Assuming the goal is to make the public API identical, lgtm

Issue #208, opened 2023-03-26
Support reading source JARs


mahesh-hegde:
summarizer can read

- compiled JARs

- `.class` files in filesystem paths

- `.java` source files in filesystem paths.

Source JARs are currently a notable omission.

This is small implementation detail - because we need to pass a `JavaFileObject` to `DocletSummarizer`. But we can subclass `SimpleJavaFileObject` to refer to a `{JarFile, ZipEntry}` pair.

__Follow up question:__

While we are at it, should default maven source download format be changed from `unpack-dependencies` (scatters all source files in package layout) to `copy-dependencies` (just copies the JARs, similar to what we do with compiled JARs)?

cah-aamir:
Does this mean at present bindings can't be generated for jar files read from local memory? I'm trying to run ffigen on a 3rd party jar file that is not available on maven and seems like it can't find any classes.

mahesh-hegde:
> Does it support generating bindings for jar files containing compiled .class files read from local memory?

Yes.

What's the error? This shouldn't happen. It's great if you can share config too.

cah-aamir:
> > Does it support generating bindings for jar files containing compiled .class files read from local memory?
> 
> Yes.
> 
> What's the error? This shouldn't happen. It's great if you can share config too.

So my jar file is in sdk folder (sdk/jackson-core-2.13.4.jar) and here's the config:
`output:
  c:
    library_name: example
    path: src/example/
  dart:
    path: lib/example.dart
    structure: single_file

class_path:
  - 'sdk/'

classes:
  - 'com.fasterxml.jackson.core.JsonFactory'`

mahesh-hegde:
This may be an instance of #147.

Will be pushing a fix soon.

Until then, try giving a dummy source path eg an empty dir.

cah-aamir:
> > Does it support generating bindings for jar files containing compiled .class files read from local memory?
> 
> Yes.
> 
> What's the error? This shouldn't happen. It's great if you can share config too.

Looks like I'm missing something in the config. 
My jar file is in sdk folder (sdk/jackson-core-2.13.4.jar) and here's the config:
```
output:
  c:
    library_name: example
    path: src/example/
  dart:
    path: lib/example.dart
    structure: single_file

class_path:
  - 'sdk/'

classes:
  - 'com.fasterxml.jackson.core.JsonFactory'
```
Error message:
```
(jnigen) INFO: ApiSummarizer.jar exists. Skipping build..
[...]

PR #220, opened 2023-03-31
JAR handling improvements

mahesh-hegde:
A refactor of summarizer's JAR file handling and class searching, which hopefully

* closes: #208 
* closes: #147 

This will be a big PR, so take time to review.

Edit: The JAR files will stay open for the duration of summarizer's execution. It was the easy way. But if it would cause a problem such as overflowing file descriptor limit per process, I would have to reconsider the approach.

### DIFF
--- a/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
+++ b/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
@@ -23,7 +23,7 @@ class JarEntryFileObject extends SimpleJavaFileObject {
   private int getEntrySize(ZipEntry entry) {
     long limit = 1024L * 1024L * 16L; // Arbitrary limit, how long can be a source file?
     long size = entry.getSize();
-    return (int) Long.max(size, limit);
+    return (int) Long.min(size, limit);
   }
 
   @Override
--- a/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
+++ b/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
@@ -9,6 +9,7 @@ import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 import javax.tools.SimpleJavaFileObject;
 
+/** Implements JavaFileObject for use by Doclet summarizer. */
 class JarEntryFileObject extends SimpleJavaFileObject {
   JarFile jarFile;
   String relativePath;
@@ -22,7 +23,7 @@ class JarEntryFileObject extends SimpleJavaFileObject {
   private int getEntrySize(ZipEntry entry) {
     long limit = 1024L * 1024L * 16L; // Arbitrary limit, how long can be a source file?
     long size = entry.getSize();
-    return size > limit ? (int) limit : (int) size;
+    return (int) Long.max(size, limit);
   }
 
   @Override
--- a/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
+++ b/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
@@ -1,0 +1,39 @@
+package com.github.dart_lang.jnigen.apisummarizer.util;
+
+import javax.tools.SimpleJavaFileObject;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.jar.JarFile;
+import java.util.zip.ZipEntry;
+
+class JarEntryFileObject extends SimpleJavaFileObject {
+    JarFile jarFile;
+    String relativePath;
+
+    protected JarEntryFileObject(JarFile jarFile, ZipEntry entry) {
+        super(URI.create(jarFile.getName() + "!" + entry.getName()), Kind.SOURCE);
+        this.jarFile = jarFile;
+        this.relativePath = entry.getName();
+    }
+
+    private int getEntrySize(ZipEntry entry) {
+        long limit = 1024L * 1024L * 16L; // Arbitrary limit, how long can be a source file?
+        long size = entry.getSize();
+        return size > limit ? (int) limit : (int) size;
+    }
+
+    @Override
+    public CharSequence getCharContent(boolean ignoreEncodingErrors) {
+        var entry = jarFile.getEntry(relativePath);
+        var out = new ByteArrayOutputStream(getEntrySize(entry));
+
+        try (var stream = jarFile.getInputStream(entry)) {
+            stream.transferTo(out);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        return out.toString(StandardCharsets.UTF_8);
+    }
+}
--- a/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
+++ b/jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
@@ -1,6 +1,5 @@
 package com.github.dart_lang.jnigen.apisummarizer.util;
 
-import javax.tools.SimpleJavaFileObject;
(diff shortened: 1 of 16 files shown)


mahesh-hegde:
Good CIs create men who forget to run their pre-push checks.

jnigen/lib/src/summary/summary.dart
```diff
@@ -86,3 +91,85 @@ class SummarizerCommand {
     return proc;
   }
 }
+
+Future<Classes?> getSummary(Config config) async {
+  setLoggingLevel(config.logLevel);
```
HosseinYousefi, line 96: You're setting the logging level, here and in `generateJniBindings`.

jnigen/lib/src/summary/summary.dart
```diff
@@ -86,3 +91,85 @@ class SummarizerCommand {
     return proc;
   }
 }
+
+Future<Classes?> getSummary(Config config) async {
```
HosseinYousefi, line 95: This can return a `Future<Classes>` instead. The code after `log.fatal` never runs and only then you're returning null.

jnigen/lib/src/generate_bindings.dart
```diff
@@ -23,83 +22,12 @@ Future<void> generateJniBindings(Config config) async {
...
-  Process process;
-  Stream<List<int>> input;
-  try {
-    process = await summarizer.runProcess();
-    input = process.stdout;
-  } on Exception catch (e) {
-    log.fatal('Cannot obtain API summary: $e');
+  final classes = await getSummary(config);
+  if (classes == null) {
+    log.fatal("Error occurred while parsing inputs.");
```
HosseinYousefi, line 27: You're already handling the errors in `getSummary`.

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/SearchUtil.java
```diff
@@ -0,0 +1,159 @@
...
+
+    var d = new File(searchPath, s);
+    if (d.exists() && d.isDirectory()) {
+      return Optional.of(recursiveListFiles(d, file -> file.getName().endsWith(".java")));
+    }
+
+    return Optional.empty();
+  }
+
+  /// find files in jar
```
HosseinYousefi, line 32: The name is descriptive enough. So no need for these comments.

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
```diff
@@ -0,0 +1,41 @@
...
+  protected JarEntryFileObject(JarFile jarFile, ZipEntry entry) {
+    super(URI.create(new File(jarFile.getName()).toURI() + "/" + entry.getName()), Kind.SOURCE);
+    this.jarFile = jarFile;
+    this.relativePath = entry.getName();
+  }
+
+  private int getEntrySize(ZipEntry entry) {
+    long limit = 1024L * 1024L * 16L; // Arbitrary limit, how long can be a source file?
+    long size = entry.getSize();
+    return (int) Long.max(size, limit);
```
HosseinYousefi, line 26: Why do we need this lower limit?

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.zip.ZipEntry;
+import javax.tools.SimpleJavaFileObject;
+
+/** Implements JavaFileObject for use by Doclet summarizer. */
+class JarEntryFileObject extends SimpleJavaFileObject {
+  JarFile jarFile;
+  String relativePath;
+
+  protected JarEntryFileObject(JarFile jarFile, ZipEntry entry) {
+    super(URI.create(new File(jarFile.getName()).toURI() + "/" + entry.getName()), Kind.SOURCE);
```
HosseinYousefi, line 18: Use a platform dependant path separator instead of `"/"`?

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/SearchUtil.java
```diff
@@ -0,0 +1,159 @@
...
+  public static Optional<List<ZipEntry>> findFilesInJar(
+      String qualifiedName, JarFile jar, String suffix) {
+    String relativePath = qualifiedName.replace(".", "/");
+    var classEntry = jar.getEntry(relativePath + suffix);
+    if (classEntry != null) {
+      return Optional.of(List.of(classEntry));
+    }
+    var dirPath = relativePath.endsWith("/") ? relativePath : relativePath + "/";
```
HosseinYousefi, line 40: Platform dependant path separators. Used in different places within this file.

HosseinYousefi (changes requested):

jnigen/lib/src/summary/summary.dart
mahesh-hegde: This is a potential entry point, eg during tests of this particular feature. If logs during that test is fine, I will remove this line.

In long term we may want to design a better logging abstraction. But I wouldn't worry about it a lot now.

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/JarEntryFileObject.java
mahesh-hegde: Sorry I should be using `Long.min`. That's an embarassing mistake.

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/SearchUtil.java
mahesh-hegde: It's inside JAR (used in `getEntry`). I am not sure I can use a platform dependent separator there.

jnigen/lib/src/summary/summary.dart
mahesh-hegde: I need to still return a Future.error. (Otherwise compiler will not know that `log.fatal` always exits).

mahesh-hegde: Wait I can make `log.fatal` return `Never`, right?

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/SearchUtil.java
```diff
@@ -29,7 +28,6 @@ public static Optional<List<File>> findFilesInPath(
     return Optional.empty();
   }
 
-  /// find files in jar
   public static Optional<List<ZipEntry>> findFilesInJar(
       String qualifiedName, JarFile jar, String suffix) {
     String relativePath = qualifiedName.replace(".", "/");
```
HosseinYousefi, line 33: also here

HosseinYousefi (approved):
LGTM! Thanks @mahesh-hegde!

jnigen/java/src/main/java/com/github/dart_lang/jnigen/apisummarizer/util/SearchUtil.java
mahesh-hegde: It seems Jar entries use / in separators regardless of the OS. Probably a backslash will work but having parity with getName() etc methods on JarEntry is a good idea.

```
jshell> var j = new JarFile(".dart_tool/jnigen/ApiSummarizer.jar")
j ==> java.util.jar.JarFile@146ba0ac

jshell> j.stream().limit(5).forEach(e -> System.out.println(e))
META-INF/
META-INF/MANIFEST.MF
com/
com/github/
com/github/dart_lang/
```

mahesh-hegde: It seems Jar entries use / in separators regardless of the OS. Probably a backslash will work but having parity with getName() etc methods on JarEntry is a good idea.

```
jshell> var j = new JarFile(".dart_tool/jnigen/ApiSummarizer.jar")
j ==> java.util.jar.JarFile@146ba0ac

jshell> j.stream().limit(5).forEach(e -> System.out.println(e))
META-INF/
META-INF/MANIFEST.MF
com/
com/github/
com/github/dart_lang/
```

PR #222, opened 2023-04-01
(An attempt to) Improve CI workflows

mahesh-hegde:
I think CI has too many jobs and too much time to run. This is an attempt to reduce that.

Marking it as draft & continuously pushing because that's the only way to ensure GH actions work.


mahesh-hegde:
@dcharkes Is it necessary to ensure build targetting android works on windows AND linux, separately?

That's usually the longest running job in my experience.

dcharkes:
> * should android build be run on each host OS? As long as the generated bindings are same, build should succeed on any host, right? Removing android builds from windows test job will reduce its runtime significantly.

Theoretically yes, but if there's anything wrong with the build setup it might fail on one host. I rather test too much than too little, stuff breaks in weird and unpredictable ways some times.

(The more we do with unit tests, the less integration tests we have to run, so maybe we can focus on that and then have a single integration test that tests a lot of different things as a smoke test.)

> * Why are is `test_with_coverage` run after `dart test`? If we are collecting coverage, isn't it sufficient to directly run `test_with_coverage`?

Correct, we could remove the `dart test`.

mahesh-hegde:
> The more we do with unit tests, the less integration tests we have to run, so maybe we can focus on that and then have a single integration test that tests a lot of different things as a smoke test.

This is true. On the other hand, only 3.5 minutes out of 20 minutes in the windows job are spent in `dart test`. Android builds are taking ~10 minutes together.

I agree that test suite itself shouldn't take 3.5-4 minutes anyway, we should optimize that in future. But in this PR I am looking at low hanging fruit.

mahesh-hegde:
However the problem with test_with_coverage is it doesn't surface all errors.

Now I recall previously a timeout error was no properly reported in `test_with_coverage`.

Since the jobs which report coverage aren't the bottleneck anyway, I think this PR can be merged. Will submit different PR if I can find more low hanging fruit.

jnigen/test/yaml_config_test.dart
```diff
@@ -1,30 +0,0 @@
-// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
dcharkes, line 1: Isn't this test testing anything over the other tests that we remove it? (Apparently we don't drop coverage...)

dcharkes (approved):
LGTM assuming this makes the bots run faster and does not drop coverage.

jnigen/test/yaml_config_test.dart
HosseinYousefi: I guess now that we use `package:cli_config`, we can rely on the tests there.

HosseinYousefi (approved):
LGTM!

jnigen/test/yaml_config_test.dart
mahesh-hegde: > Isn't this test testing anything over the other tests that we remove it? (Apparently we don't drop coverage...)

Apparently because we have a test which compares the same YAML with programmatic config line by line, and the latter is used in integration test!

PR #231, opened 2023-04-05
Add ability to use source dependencies from Gradle

mahesh-hegde:
* Add ability to use source dependencies of android gradle project (Using a stub again, sorry :))

* Add example of this functionality using `androidx.emoji2.text.EmojiCompat` class.

* Provide useful error message when `build.gradle` is left around with the stub function. (However, flutter tool may suggest an irrelevant fix which may cause this to be ignored by not-so-careful reader.)


HosseinYousefi (approved):
LGTM! Thanks @mahesh-hegde 🚀

PR #234, opened 2023-04-06
Rename add_gradle_deps to add_gradle_jars

mahesh-hegde:
This is a breaking change though, so someone looking at examples here and trying this in jnigen 0.3.0 may be confused.

Maybe we should just keep this here and merge during v0.4.0 release.

### DIFF
--- a/jnigen/example/in_app_java/jnigen.yaml
+++ b/jnigen/example/in_app_java/jnigen.yaml
@@ -1,5 +1,5 @@
 android_sdk_config:
-  add_gradle_deps: true
+  add_gradle_jars: true
   add_gradle_sources: true
 
 output:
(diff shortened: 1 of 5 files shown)


HosseinYousefi:
We could introduce some flag that combines the two like `add_gradle_deps` that can be `all`, `none`, `sources_only` and `jars_only`. I think I would mostly want to use `add_gradle_deps: 'all'`. Naming could be improved.

HosseinYousefi:
@mahesh-hegde Let's not do this for this release. Instead I'll open an issue to discuss a good naming strategy and maybe do it for 0.5.0.

mahesh-hegde:
Alright.

Issue #141, opened 2022-11-18
Support Wildcards (in generics)


HosseinYousefi:

mahesh-hegde:
Related Generics issues:

#145 
#144 
#159 

It might be better to map the JLS generics to Dart generics upfront for once, in form of a specification.

mahesh-hegde:
I face this problem when implementing parsing of wildcard in `asm` (`TypeUsageSignatureVisitor`)[^problem]. So I am looking for a way to solve this if possible.

What are our options here?

1. `? extends QualifiedName` will map to `QualifiedName` and `? super QualifiedName` will map to `Object`. This is probably easiest to implement, and with current implementation of `TypeSignatureVisitor` in summarizer, this is what happens anyway.

2. `? extends QualifiedName` will create a new type variable `W1 extends QualifiedName`. `? super QualifiedName` is still `Object`.

I need to study the generics spec, but I would like to listen from those who already have. :)

[^problem]: Map<K, V> uses many `? extends K` kind of stuff. Previously `+K` would get parsed as `K` but if we want to parse it as `? extends K`, we should handle `Kind.WILDCARD` properly in the generator.

PR #236, opened 2023-04-07
Type inference for generics

### LINKED ISSUE #145: Type "inference" for generics
### Simple example

Let's take this add method:

```java
// In class Test
static public <T> T transfer(ArrayList<T> l, Stack<T> s) {
  while (!s.empty()) {
    l.add(s.pop());
  }
  return l.get(0); // just so it's not void
}
```

Right now, calling this in Dart is like this:

```dart
Test.transfer(Integer.type, intList, intStack);
```

Even though in Java, we're able to do

```java
Test.transfer(intList, intStack);
```

and have it automatically infer the type parameter `T`.

We could try doing the same in Dart. Instead of passing `T` as the first ordered argument, we can pass it as an (in this case) optional type parameter.

```dart
static T transfer<T>(
  ArrayList<T> l,
  Stack<T> s, {
  JObjType<T>? $T,
}) {
  // ...
}
```

In the body of the function, we will have to have something like:

```dart
$T ??= (l.$type as $ArrayListType<T>).$T;
return $T.fromRef(...);
```

Also it's useful to have assertions making sure `l.$type.$T` and `s.$type.$T` are the exact same. This is because in Dart we could pass `ArrayList<Integer>` and `Stack<Number>`. `T` will be `Number` but we know that a `Number` cannot be added to an `ArrayList<Integer>`!

```dart
assert((l.$type as $ArrayListType<T>).$T == (s.$type as $StackType<T>).$T, '...');
```

The type parameter in this case looks like so:

```mermaid
graph TD;
  ArrayList--$T-->T;
  Stack--$T-->T;
```

### Higher depths!

But they can be arbitrarily deep, for example if instead of `ArrayList<T>` we have `ArrayList<HashMap<String, T>>`:

```mermaid
graph TD;
  ArrayList--$T-->HashMap;
  HashMap--$K-->String;
  HashMap--$V-->T;
  Stack--$T-->T;
```

The generated code also gets more complicated:

```dart
assert((l.$type as $ArrayListType<HashMap<JString, T>>).$T.$V == (s.$type as $StackType<T>).$T, '...');
```

We'd also need `k - 1` equality assertions for `k` arguments that have `T` type parameter. Of course, this gets more complicated when we have multiple type parameters.

### Different variants

`transfer` method in the example above can only transfer elements of `Stack<T>` to an `ArrayList<T>`. We want to make this method more generic, so that we can transfer any `Stack<U>` to an `ArrayList<T>` where `U <: T`. Let's change the signature to allow this:

```java
// In class Test
static public <T> T transfer(ArrayList<T> l, Stack<? extends T> s) {
  while (!s.empty()) {
    l.add(s.pop());
  }
  return l.get(0);
}
```

Now the assertion in Dart would have to be changed as well:

```dart
assert((l.$type as $ArrayListType<T>).$T.isSuperOf((s.$type as $StackType<T>).$T), '...');
```

`? super T` would be similar to `? extends T`.

### Lowest common ancestor type

Now let's say that we don't have any type `T` directly.

```java
static public <T> T randomFirst(ArrayList<? extends T> l, Stack<? extends T> s) {
  if (random() % 2 == 0) {
    return l.get(0);
  }
  return s.get(0);
}
```

In this case `$T` should be the lowest common ancesstor of `s.$type.$T` and `t.$type.$T`. Take these classes:

```mermaid
graph TD;
  java.lang.Object--subclass-->A
  A--subclass-->B
  B--subclass-->C
  A--subclass-->D
  D--subclass-->E
```

Here `Test.randomFirst(listOfC, listOfE)` should have a type of `A`.

### Supers and extends

What about a combination of `super` and `extends`?

```java
static public <T> T setFirst(ArrayList<? super T> l, Stack<? extends T> s) {
  l.set(0, s.get(0));
  return s.get(0);
}
```

This should only work when `l.$type.$T` is a supertype of `s.$type.$T`.

### Conclusion

As you can see, there are many cases that need to be thoroughly tested.

### Performance

The type graph is not going to be super deep. So naive implementation does the job in most cases.

We're using the operations of type `is_ancestor?` and `lowest_common_ancestor` a lot. So keeping not only the supertypes but also jumps with powers of two (parent, grandparent, grandparent of grandparent, ...) can improve the performance in cases where we have many layers of subclass/superclasses.

HosseinYousefi:
Closes #145. Doesn't support wildcards yet, will be addressed in #141.


jnigen/example/in_app_java/lib/android_utils.dart
```diff
@@ -46,8 +46,12 @@ class AndroidUtils extends jni.JObject {
 
   /// from: static public void showToast(android.app.Activity mainActivity, java.lang.CharSequence text, int duration)
   static void showToast(
-          jni.JObject mainActivity, jni.JObject text, int duration) =>
-      _showToast(mainActivity.reference, text.reference, duration).check();
+    jni.JObject mainActivity,
+    jni.JObject text,
+    int duration,
```
dcharkes, line 51: Can we land this last comma format change separately? It would make it easier to review.

(Sidenode, I indeed prefer having a trailing comma, it uses more lines but makes code easier to read in general.)

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -63,6 +73,31 @@ String _newLine({int depth = 0}) {
   return '\n${'  ' * depth}';
 }
 
+/// Merges two maps by adding all the values of one to the other.
```
dcharkes, line 76: Nit: you're making a new outer map rather than adding the values of one to the other.

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -1131,3 +1225,137 @@ class _JValueWrapper extends TypeVisitor<String> {
...
+    final result = <String, List<_PrependingBuffer>>{};
+    final prefix = resolver?.resolvePrefix(node.binaryName) ?? '';
+    final typeClass =
+        '$prefix$_typeClassPrefix${node.classDecl.finalName}$_typeClassSuffix';
+    for (var i = 0; i < node.params.length; ++i) {
+      final typeParam = node.classDecl.allTypeParams[i + offset].name;
+      final exprs = node.params[i].accept(this);
+      for (final expr in exprs.entries) {
+        for (final buffer in expr.value) {
+          // [_PrependingBuffer] adds the correct number of openning paranthesis
```
dcharkes, line 1336: parenthesis

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -1131,3 +1225,137 @@ class _JValueWrapper extends TypeVisitor<String> {
     return '$_jni.JValue${node.name.capitalize()}($param)';
   }
 }
+
+/// A [StringBuffer] that prepends instead of appending on write.
```
dcharkes, line 1229: This feels a bit of a wacky abstraction. Prepending and adding parenthesis feel like two orthogonal concerns.

I gather string concatenation or normal StringBuffer didn't suffice.

Should this be a reusable abstraction in its own file? In utils? Or are we only going to is it in this dart_generator?

jnigen/lib/src/elements/elements.dart
```diff
@@ -97,6 +97,12 @@ class ClassDecl extends ClassMember implements Element<ClassDecl> {
 
   String get internalName => binaryName.replaceAll(".", "/");
 
+  /// The number of super classes this type has.
+  ///
+  /// Will be populated by [Linker].
```
dcharkes, line 102: Will be populated by -> Populated by

(e.g. the comment should be true when reading it in arbitrary points in the timeline of our pipeline.)

Idem below.

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
```diff
@@ -59,7 +59,7 @@ import "../../../../_init.dart";
 /// instances.
 ///@author Tatu Saloranta
 class JsonFactory extends jni.JObject {
-  late final jni.JObjType? _$type;
+  jni.JObjType? _$type;
```
dcharkes, line 62: Why is this no longer final? Does the time change over time?

jni/test/type_test.dart
```diff
@@ -0,0 +1,24 @@
...
+  if (!Platform.isAndroid) {
+    try {
+      Jni.spawn(dylibDir: "build/jni_libs", jvmOptions: ["-Xmx128m"]);
+    } on JvmExistsException catch (_) {
+      // TODO(#51): Support destroying and reinstantiating JVM.
+    }
+  }
+
+  test('commonType', () {
+    expect(commonType([JObject.type, JObject.type]), JObject.type);
```
dcharkes, line 20: Yay for more unit tests!

jni/test/type_test.dart
```diff
@@ -0,0 +1,24 @@
...
+    expect(commonType([JString.type, JString.type]), JString.type);
+    expect(commonType([JString.type, JArray.type(JLong.type)]), JObject.type);
```
dcharkes, line 22: Maybe add unit tests that don't end in JObject.type or the type itself.

jnigen/lib/src/bindings/dart_generator.dart
HosseinYousefi: Both of them do suffice but they're going to be slower. For one, we have to prepend instead of appending, doing it using a normal `StringBuffer` would make it `O(n)`, and we would want to keep the count of the closing parenthesis to be able to open them. 

This is not going to be reusable in any other context. Just for this place.

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
HosseinYousefi: It's actually a mistake to do `somethingLate ??= something;` because `somethingLate` would have to be evaluated first to check for nullity but it's never been initialized since it's `late`. It's never changed, it's just caching `$type`.

jni/lib/src/jarray.dart
```diff
@@ -16,6 +16,12 @@ class JArrayType<T> extends JObjType<JArray<T>> {
 
   @override
   JArray<T> fromRef(Pointer<Void> ref) => JArray.fromRef(elementType, ref);
+
+  @override
+  JObjType get superType => const JObjectType();
```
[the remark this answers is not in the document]
HosseinYousefi: Later when we have interfaces, this logic might be different, as we can have a superType and many implemented interfaces. But for now superType suffices.

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
dcharkes: Correct, and you don't want it `final` because you want to initialize it at some point on demand instead of in the constructor.

I'm not sure what is a nice pattern for that. This doesn't communicate it doesn't change after it's initialized. Can't we just have an initialization expression instead?

```dart
  @override
  late final jni.JObjType $type = type;
```

jni/lib/src/jarray.dart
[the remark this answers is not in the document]
dcharkes: Please add a reference to an open bug for the places where we would likely need to change things.

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
HosseinYousefi: I'm not sure if that's a good idea because even if we never use `$type`, it's still going to be initialized, right?

jnigen/lib/src/bindings/dart_generator.dart
HosseinYousefi: I will improve the wordings, these are "multimaps", i.e. maps with a List as their values, I'm merging these lists, and I'm not creating new lists, I reuse the larger one to get a logarithmic complexity instead of a quadratic one.

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
dcharkes: No, `late final`s are only initialized on demand.

(See discussion on https://dart-review.googlesource.com/c/sdk/+/286782/3..5/pkg/vm/lib/transformations/ffi/finalizable.dart#b267 )

jnigen/example/in_app_java/lib/android_utils.dart
HosseinYousefi: The reason I added this is because there may or may not be a `{...}` for the named (type) parameters. I could add some `if` statement to only add the comma if we have type parameters.

jnigen/test/jackson_core_test/third_party/lib/com/fasterxml/jackson/core/JsonFactory.dart
```diff
@@ -191,9 +199,12 @@ class JsonFactory extends jni.JObject {
   /// Constructor used by JsonFactoryBuilder for instantiation.
   ///@param b Builder that contains settings to use
   ///@since 2.10
-  JsonFactory.ctor3(jni.JObject b)
-      : super.fromRef(jniAccessors
-            .newObjectWithArgs(_classRef, _id_ctor3, [b.reference]).object);
+  factory JsonFactory.ctor3(
+    jni.JObject b,
+  ) {
```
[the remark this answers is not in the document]
HosseinYousefi: Makes the generator code extra complicated.

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -502,11 +549,11 @@ class _TypeClass {
 
 /// Generates the type class.
 class _TypeClassGenerator extends TypeVisitor<_TypeClass> {
+  _TypeClassGenerator(this.resolver, {this.isConst = true});
```
dcharkes, line 552: nit: I think it's more common to have fields first, then constructors/factories, then static methods, and then normal methods.

But, I might be projecting from the C++ code style. (I don't see any mention of this in Dart style.)

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -1131,3 +1231,141 @@ class _JValueWrapper extends TypeVisitor<String> {
...
+/// For example:
+/// ```dart
+/// final exprBuilder = ReverseExpressionBuilder();
+/// exprBuilder.write(' as A).first');
+/// exprBuilder.write(' as B).second');
+/// exprBuilder.write('third');
+/// exprBuilder.toString(); // ((third as B).second as A).first
+/// ```
+@visibleForTesting
+class ReverseExpressionBuilder {
```
dcharkes, line 1250: This still feels like a weird abstraction to me. I would almost prefer a Dart AST instead, but that would make it horribly verbose.

I'm happy to just land this.

jnigen/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.4.0
```
dcharkes, line 1: pubspec.yaml should also be 0.4.0

If we're not yet ready to release 0.4.0, it should be 0.4.0-dev. Then the first (and only) commit that marks it 0.4.0 should be tagged v0.4.0 and be released as 0.4.0. And we can then automate releasing with a GitHub bot (as here: https://github.com/dart-lang/tools/actions/workflows/publish.yaml)

dcharkes (approved):
Still LGTM!

jni/lib/src/jarray.dart
```diff
@@ -16,6 +16,22 @@ class JArrayType<T> extends JObjType<JArray<T>> {
...
+
+  @override
+  final int superCount = 1;
+
+  @override
+  int get hashCode => Object.hash(JArrayType, elementType);
+
+  @override
+  bool operator ==(Object other) {
+    return other.runtimeType == JArrayType &&
```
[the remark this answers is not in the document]
HosseinYousefi: Yes!

Issue #149, opened 2022-11-23
Remove `ffigen` "patch"


HosseinYousefi:
We're currently using a patch of `package:ffigen`, that adds certain extensions. This could be done in a separate file and use the parse methods exposed which allows us to use the latest version of `package:ffigen` without any problems.

mahesh-hegde:
I am trying this. But it seems `ffigen` only exports a handful of types - `Config`, `Library`, `parse()`. We need various `Binding` subtypes to generate extensions.

It's always possible to write our own using C++/Clang API or more painfully, libclang generated using ffigen. But I am looking for a way to use ffigen itself.

We can depend on `src/**` and probably pin the exact version. That's better than write own script.

Issue #85, opened 2022-10-02
Use mutex around method / field lookup


mahesh-hegde:
In current state, loading fields / methods can cause race condition if same MethodID or class is loaded from 2 isolates, and not initialized at the time of checking. It's rare but possible.

Option 1. Leave as it is, it's rare case and at worst may leak a reference.

Option 2: Have a common mutex for lookups.
```
inline load_class(jclass *cls, char *name) {
    if (*cls== NULL) {
        //grab common mutex using platform-specific primitive
        if (*cls== NULL) { // double check, because by the time lock is obtained, it may be initialized by the time we get lock.
           *cls = findClass(name);
        }
        //release mutex
    }
}
```

Option 3: Same as the above but Instead of common mutex, have a mutex per class variable.

cc: @dcharkes @liamappelbe

dcharkes:
I think option 1 is fine for the time being.

But we should add a section to the readme detailing that multiple isolates is only safe if the isolates are not racing in the way you describe.

mahesh-hegde:
I suppose this is changing one or two inline functions / macros, so can be marked for v0.2 or v0.3.

Issue #122, opened 2022-11-04
Singleton JVM


dcharkes:
We only support singleton interaction with the JVM (unless we address https://github.com/dart-lang/jnigen/issues/121).

We should

- [ ] rename Jni.spawn to Jni.singleton (or a better name)
- [ ] Add a mutex around JVM initialization
- [ ] Add a mutex around any other initialization https://github.com/dart-lang/jnigen/issues/85

mahesh-hegde:
Actually creation of multiple JVMs is not supported even in the standard JDK.

you get JNI_EEXIST if you try to spawn a second JVM.

So I don't think we need a lock here. Just have to catch JNI_EEXIST.

`spawn()` can remain. It doesn't return JVM.

HosseinYousefi:
Closing then.

PR #246, opened 2023-04-15
Remove FFI patch and depend on a pinned ffigen instead

mahesh-hegde:
Quite a large PR. When draft status is unmarked, should close the following issues.

* closes: #149 
* closes: #85 
* Maybe #122 
    * As I explained elsewhere, JDK itself raises JNI_EEXIST if we try to create multiple JVMs, so I am just catching that, and sleeping 1s so that entire JNI initializes. In real (non-test) use case, one isolate should spawn JVM, just like how one thread initializes locks in pthreads or other concurrent programming.

Although not explicitly filled as issues

* Do exception checking in the methods of `JniGlobalEnv`
* Return status code from `Jni.spawn` and throw it as appropriate exception in Dart
* Add few more tests to package:jni
* Copy paste all package:jni tests to jni/example, which allows us to run them as integration test.
* Complete few more methods not renamed in #125 - Eg: `findJniClass -> findJClass`

This is in draft because I have to do some more small changes (also relying on CI to test on all platforms), But feel free to review it until then.

### Edit: TODO for this PR

- [ ] Wrap GlobalJNIEnv in a class, and cache result of `asFunction()`. Otherwise we will be leaking FFI trampolines. 🙈 

- [ ] Figure out how to (and whether to) expose local `JNIEnv` and `JavaVM` as separate API.
    - [ ] If the answer to above question is yes, add tests on local `JNIEnv` and `JavaVM`.

- [ ] Investigate static initialization of mutex / critical section on windows. If it's possible, we can do proper locking of `spawnJvm`.

- [x] What's up with the failing test


mahesh-hegde:
The test that fails on `beta` but passes on `stable` is only indentation differences after `dart format`.

Is it possible something upstream changed at beta? No dart binding was modified in this PR.

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -1119,7 +1119,7 @@ class _MethodGenerator extends Visitor<Method, void> {
     $callExpr;
     final \$o = $_jPointer.fromAddress(await \$p.first);
     final \$k = $returnTypeClass.getClass().reference;
-    if ($_jni.Jni.env.IsInstanceOf(\$o, \$k) == 0) {
+    if ($_jni.Jni.env.IsInstanceOf(\$o, \$k) == false) {
```
HosseinYousefi, line 1122: now that it's bool, generate `if (!$_jni.Jni.env.IsInstanceOf(\$o, \$k)) {` instead

jni/example/integration_test/on_device_jni_test.dart
HosseinYousefi, line 1: The fact that this is a copy pasted file means that any change in the tests has to be replicated here. Is there a better way to do this?

jni/src/dartjni.h
```diff
@@ -195,7 +269,9 @@ static inline void load_method(jclass cls,
                                const char* name,
                                const char* sig) {
   if (*res == NULL) {
+    acquire_lock(&jni->locks.methodLoadingLock);
     *res = (*jniEnv)->GetMethodID(jniEnv, cls, name, sig);
```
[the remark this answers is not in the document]
mahesh-hegde: I cherry picked this commit from a branch I made in december. All this while I was thinking I am doing double checked locking, while I am actually not 🤡

jni/example/integration_test/on_device_jni_test.dart
mahesh-hegde: Idk.

If we bend the idioms we can put actual tests in lib/src/ and call them from both places. But sure this will be frowned upon.

dcharkes: How about a relative path import from one to the other?

mahesh-hegde:
regarding the failing test:

It appears the beta sdk does formatting differently, so the test is failing.

I confirmed it by pushing an empty commit to another draft PR, which also fails now.

HosseinYousefi:
> I confirmed it by pushing an empty commit to another draft PR, which also fails now.

Maybe let's diff them without regarding to the spacing `--ignore-space-change` for now.

jni/example/integration_test/on_device_jni_test.dart
mahesh-hegde: It seems to be working. Hope `pub` doesn't find faults with that.

HosseinYousefi:
> Still failing. I would later check with beta SDK locally.

Well, it seems like the simple_package_test is failing, changing all the diffs to ignore spaces would solve the issue.

mahesh-hegde:
I am not done yet!

HosseinYousefi:
@mahesh-hegde I approved this PR with only a single nit comment, would you want to do the TODOs in another PR and merge this?

mahesh-hegde:
Coverage decreased because I forgot to add ignore_for_file on few files. I will fix that and merge.

HosseinYousefi:
> I am not done yet!

Please keep it as draft if meaningful work is remaining.

mahesh-hegde:
I apologize. GitHub was not refreshing the commits I was pushing so I
unmarked draft to see what happens.

        Mahesh

PR #253, opened 2023-04-18
Fix summary generation tests

mahesh-hegde:
There was an order dependency in summary_generation_tests which depended on class files from source dir instead of temp dir, thus would fail if no previous test had produced those files.

Was able to catch this only because tried to run all tests in a pristine linux environment. (CI machines have much less cores, which means there isn't much concurrency there).

From now on, making it a resolution to not compile anything into source dirs.


HosseinYousefi (approved):
LGTM! Thanks, @mahesh-hegde.

PR #257, opened 2023-04-20
Bunch of small testsuite improvements

mahesh-hegde:
The commit messages should list it all :)

If you don't like any of the change, I can usually revert that individual commit.

- closes: #255 
- closes: #247 
- half-completes: #248
    - After either speeding up the bottleneck in `bindings_test` or marking it as large test, I expect `dart test -x large_test,summarizer_test` to complete much faster than entire testsuite, which will be useful in dev.


HosseinYousefi (approved):
LGTM 🚀 Thanks @mahesh-hegde!

PR #259, opened 2023-04-21
Add FAQ entry about references

mahesh-hegde:


README.md
```diff
@@ -188,6 +188,14 @@ If the classes are in JAR file, make sure to provide path to JAR file itself, an
...
 If the errors are similar to `symbol not found`, ensure all dependencies of the source are available. If such dependency is compiled, it can be included in `class_path`.
 
+#### How are classes mapped into bindings?
+Each Java class generates a subclass of `JObject` class, which wraps a `jobject` reference in JNI. Nested classes use `_` as separator, `Example.NestedClass` will be mapped to `Example_NestedClass`.
+
+#### Does `JObject` hold a local or global reference? Does it need to be manually deleted?
+Each Java object returned into Dart creates a JNI global reference. Reference deletion is taken care of by `NativeFinalizer` and that's usually sufficient.
+
+It's a good practice to keep the interface between languages sparse. However, if there's a need to create several references (Eg: in a loop), you can use FFI Arena mechanism (`using` function) and `deletedIn` method, or manually delete the object using `delete` method.
+
```
dcharkes, line 198: Should we document the 64k limit, and the symptoms for what it looks like when one hits that limit?

dcharkes (approved):

README.md
mahesh-hegde: Symptom is crashing :)

About 64K limit, is that official or just internal knowledge?

dcharkes: Quick search revealed this: https://android.googlesource.com/platform/dalvik.git/+/android-4.3_r2/vm/Jni.cpp#250

HosseinYousefi (approved):

README.md
mahesh-hegde: So should this be documented or not? Your call.

dcharkes: I would add something about what the crash looks like, so that it will help users if they run into this.

mahesh-hegde: It will explicitly state JNI global reference table overflowed. Should an exact message be included?

mahesh-hegde: It's more complicated than that. Sometimes the app may crash without output. In general we can't rely on an output being produced to flutter console.

Issue #199, opened 2023-03-10
Unable to generate files from jnigen


thisisamank:
I am trying to setup `jnigen` and run the basic sample project.
edit: I have tried this on jdk 18,17 and 11.
On running  `dart run jnigen --config jnigen.yaml`
I am getting this
``` bash
(jnigen) INFO: Building ApiSummarizer component. This might take some time. 
The build will be cached for subsequent runs

(jnigen) INFO: execute mvn --batch-mode --update-snapshots -f /Users/thisisamank/.pub-cache/hosted/pub.dev/jnigen-0.3.0/java/pom.xml assembly:assembly
[INFO] Scanning for projects...
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for 'com.github.dart_lang.jnigen:ApiSummarizer:jar:0.0.1-SNAPSHOT'
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-assembly-plugin is missing. @ line 53, column 21
[WARNING] 
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING] 
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING] 
[INFO] Downloading from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-assembly-plugin/maven-metadata.xml
[INFO] Downloaded from central: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-assembly-plugin/maven-metadata.xml (1.3 kB at 2.2 kB/s)
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  0.800 s
[INFO] Finished at: 2023-03-10T22:35:10+05:30
[INFO] ------------------------------------------------------------------------
[ERROR] Could not find goal 'assembly' in plugin org.apache.maven.plugins:maven-assembly-plugin:3.5.0 among available goals help, single -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the '-e' switch
[ERROR] Re-run Maven using the '-X' switch to enable verbose output
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoNotFoundException


maven exited with 1
Unhandled exception:
[...]

PR #261, opened 2023-04-24
Support Maven version 4

alex-james-dev:
Currently building ApiSummarizer.jar fails if the user is running Maven version 4.x.  This pull request resolves this issue.

Closes: #199

### DIFF
--- a/jnigen/java/pom.xml
+++ b/jnigen/java/pom.xml
@@ -54,7 +54,9 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
                 <configuration>
-                    <descriptorId>jar-with-dependencies</descriptorId>
+                    <descriptorRefs>
+                        <descriptorRef>jar-with-dependencies</descriptorRef>
+                    </descriptorRefs>
                     <appendAssemblyId>false</appendAssemblyId>
                     <archive>
                         <manifest>
--- a/jnigen/java/pom.xml
+++ b/jnigen/java/pom.xml
@@ -44,7 +44,7 @@
     <build>
 		<directory>${buildDir}/target</directory>
         <finalName>ApiSummarizer</finalName>
-        <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
+        <pluginManagement><!-- Lock down plugins versions to avoid using Maven defaults -->
             <plugins>
                 <plugin>
                     <artifactId>maven-clean-plugin</artifactId>
--- a/jnigen/java/pom.xml
+++ b/jnigen/java/pom.xml
@@ -44,6 +44,22 @@
     <build>
 		<directory>${buildDir}/target</directory>
         <finalName>ApiSummarizer</finalName>
+        <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
+            <plugins>
+                <plugin>
+                    <artifactId>maven-clean-plugin</artifactId>
+                    <version>3.2.0</version>
+                </plugin>
+                <plugin>
+                    <artifactId>maven-resources-plugin</artifactId>
+                    <version>3.3.1</version>
+                </plugin>
+                <plugin>
+                    <artifactId>maven-jar-plugin</artifactId>
+                    <version>3.3.0</version>
+                </plugin>
+            </plugins>
+        </pluginManagement>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/jnigen/java/pom.xml
+++ b/jnigen/java/pom.xml
@@ -68,7 +68,7 @@
                 <executions>
                     <execution>
                         <id>make-assembly</id>
-                        <phase>package</phase>
+                        <phase>compile</phase>
                         <goals>
                             <goal>single</goal>
                         </goals>
--- a/jnigen/java/pom.xml
+++ b/jnigen/java/pom.xml
@@ -53,6 +53,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.5.0</version>
                 <configuration>
                     <descriptorRefs>
                         <descriptorRef>jar-with-dependencies</descriptorRef>
--- a/jnigen/java/pom.xml
+++ b/jnigen/java/pom.xml
@@ -65,6 +65,15 @@
                         </manifest>
                     </archive>
                 </configuration>
+                <executions>
+                    <execution>
+                        <id>make-assembly</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
--- a/jnigen/lib/src/tools/build_summarizer.dart
+++ b/jnigen/lib/src/tools/build_summarizer.dart
@@ -33,7 +33,7 @@ Future<void> buildApiSummarizer() async {
     '--batch-mode',
     '--update-snapshots',
     '-f',
-    pom.toFilePath()
+    pom.toFilePath(),
   ];
   log.info('execute mvn ${mvnArgs.join(" ")}');
   try {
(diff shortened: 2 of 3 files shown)


jnigen/lib/src/tools/build_summarizer.dart
```diff
@@ -29,11 +29,11 @@ Future<void> buildApiSummarizer() async {
   final pom = pkg.resolve('java/pom.xml');
   await Directory(toolPath).create(recursive: true);
   final mvnArgs = [
+    'package',
     '--batch-mode',
     '--update-snapshots',
     '-f',
-    pom.toFilePath(),
-    'assembly:assembly'
+    pom.toFilePath()
```
mahesh-hegde, line 36: It's a best practice to add trailing comma, after last element of any multi-line list.

jnigen/lib/src/tools/build_summarizer.dart
```diff
@@ -29,11 +29,11 @@ Future<void> buildApiSummarizer() async {
   final pom = pkg.resolve('java/pom.xml');
   await Directory(toolPath).create(recursive: true);
   final mvnArgs = [
+    'package',
```
mahesh-hegde, line 32: `package` or `compile` should not make much difference here, because every merged PR will have it's Java tests passed in CI.

mahesh-hegde (approved):
Good work.

I think there was one warning in the output you posted. Try to remove that too if possible.

jnigen/lib/src/tools/build_summarizer.dart
alex-james-dev: TIL! 

I've pushed a commit adding the trailing comma :smiley:

jnigen/lib/src/tools/build_summarizer.dart
alex-james-dev: Interesting, that's good to know about the CI testing, I'm happy with either `package` or `compile`.

If we don't need to run the Java tests due to CI, then maybe `compile` is better because on my machine `compile` is faster taking about 10s whereas `package` took about 17s. 

I'll make a commit which uses `compile` instead of `package` as it looks like the best option because it is faster and we don't need to run the tests.

alex-james-dev:
>Good work.

Thanks @mahesh-hegde :smile: Your instructions were very helpful.

> I think there was one warning in the output you posted. Try to remove that too if possible.

You're right, I think the changes have introduced a new Maven warning:

```Console
[WARNING] Version not locked for default bindings plugins [maven-resources-plugin, maven-jar-plugin], you should define versions in pluginManagement section of your pom.xml or parent
```

Interestingly, it looks like this warning only occurs on Maven 4.x.

I've pushed a commit which resolves this warning.

# Other Maven Warnings
There are still three Maven warnings but I don't think we can resolve them.

``` Console
[INFO] --- assembly:3.5.0:single (make-assembly) @ ApiSummarizer ---
[WARNING] Parameter 'localRepository' is deprecated core expression; Avoid use of ArtifactRepository type. If you need access to local repository, switch to '${repositorySystemSession}' expression and get LRM from it instead.
```
I think the above warning is an internal warning from Maven Assembly Plugin, so we can't fix this warning.

```Console
[WARNING] Configuration option 'appendAssemblyId' is set to false.
Instead of attaching the assembly file: /home/alexander/Documents/Software/GSoC 2023/Maven Bug/jnigen/jnigen/java/target/ApiSummarizer.jar, it will become the file for main project artifact.
NOTE: If multiple descriptors or descriptor-formats are provided for this project, the value of this file will be non-deterministic!
[WARNING] Replacing pre-existing project main-artifact file: /home/alexander/Documents/Software/GSoC 2023/Maven Bug/jnigen/jnigen/java/target/classes
with assembly file: /home/alexander/Documents/Software/GSoC 2023/Maven Bug/jnigen/jnigen/java/target/ApiSummarizer.jar
```
The above two Maven warnings occur because we want the final jar to be named `ApiSummarizer.jar` instead of `ApiSummarizer-jar-with-dependencies.jar`. The only way I've found to remove these warnings is to set appendAssemblyId to true (default value) but that results in the jar being called `ApiSummarizer-jar-with-dependencies.jar` which we don't want. I couldn't find a way to resolve the warnings while preserving the jar name `ApiSummarizer.jar`.

alex-james-dev:
I searched the codebase for any other occurrences of `mvn assembly:assembly`, I only found one other occurrence in the [README for ApiSummarizer.](https://github.com/dart-lang/jnigen/blob/6b1071386928fc3261a5b7357ff5bc761f1ddec1/jnigen/java/README.md?plain=1#L11):

```
To build the jar manually, run `mvn assembly:assembly` in project root.
```

Once we have decided between `mvn compile` and `mvn package`, I can update these instructions to either:

 ```
To build the jar manually, run `mvn compile` in project root.
```

or

```
To build the jar manually, run `mvn package` in project root.

jnigen/java/pom.xml
```diff
@@ -44,6 +44,22 @@
     <build>
 		<directory>${buildDir}/target</directory>
         <finalName>ApiSummarizer</finalName>
+        <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
```
mahesh-hegde, line 47: There's no parent POM here, 🤔 

Did you take this from somewhere like stack overflow post? In that case I usually make it a habit to link to the answer.

mahesh-hegde:
> To build the jar manually, run `mvn compile` in project root.

I think this is better. As you mentioned, compile time is less.

Or maybe

```
To build the jar manually, run `mvn compile` in project root. To build the jar and run the tests as well, run `mvn package`.
```

This README is only for developers anyway. So add the following point.

jnigen/java/pom.xml
alex-james-dev: > Did you take this from somewhere like stack overflow post? In that case I usually make it a habit to link to the answer.

Yeah you're right I didn't write that comment :sweat_smile: I copied it from the Maven [Getting Started Guide](https://maven.apache.org/guides/getting-started/).

As you mentioned, we don't have a parent pom so perhaps we should edit the comment to:

```suggestion
        <pluginManagement><!-- Lock down plugins versions to avoid using Maven defaults -->
```

Are we likely to add a parent pom in the future? If so we could edit the comment to acknowledge that we don't have a parent pom currently but if we do create a parent pom we should move the plugin management code to it:

```suggestion
        <pluginManagement><!-- Lock down plugins versions to avoid using Maven defaults (if we create a parent pom this should be moved to parent pom) -->
```
Or is the comment actually providing any real benefit? We could consider just removing it entirely:
```suggestion
        <pluginManagement>
```

What do you think is the best option for the comment?

mahesh-hegde: Just remove the comment. Version pinning is a very common practice.

I don't think `summarizer` will ever be complex enough to demand multiple poms.

alex-james-dev: Sounds good to me! So I'll push a commit making the following change:

```suggestion
        <pluginManagement><!-- Lock down plugins versions to avoid using Maven defaults -->
```

alex-james-dev:
> ```
> To build the jar manually, run `mvn compile` in project root. To build the jar and run the tests as well, run `mvn package`.
> ```

I agree it's a good idea to also include the instructions to run the tests, I'll definitely add a commit to include this.

Because we are building ApiSummarizer.jar during the compile phase using Maven Assembly Plugin, running `mvn package` throws an error because the jar already exists when Maven gets to the package phase. So to build the jar and run the tests as well we want to run `mvn test`.

For testing I manually ran `mvn test` in `jnigen/java` using Maven 4.0.0-alpha-5 and Maven 3.9.1, for both versions the jar was built successfully and the tests were run.

I'll push a commit changing the instructions to:

```Console
To build the jar manually, run `mvn compile` in project root. To build the jar and run the tests as well, run `mvn test`.
```

mahesh-hegde (approved):
Approved.

A note to @alex-james-dev 

* [Merge branch 'dart-lang:main' into maven_bug](https://github.com/dart-lang/jnigen/pull/261/commits/ea44252889df9043cdc1281cbaa6b294a60f609e) this commit is superflous. A merge or rebase should not be required unless there's a conflict. (Which github will tell you, in the same UI widget which shows CI checks and approval status).

In general, understanding how git works, as well as few commands beyond "normie" commit and push, (revert, reset, stash, cherry-pick, diff) is very useful. I don't claim to understand git myself, but I am making an effort :)

HosseinYousefi (approved):
Thanks @alex-james-dev!

PR #271, opened 2023-05-03
Boxed types + Map + Set + List

### LINKED ISSUE #651: Not releasing the `_classRef` inside the classes.
We currently generate `static final _classRef = jniAccessors.getClassOf(...)`. We don't release these classes. We should instead use `findJClass` for them to be `Finalizable`.

### LINKED ISSUE #711: Dart equivalent of commonly used Java types
We currently have `JString` which represents `java.lang.String`, and it's not a generated class. This is nice because we can add custom methods like `toDartString` and make the developer experience better. The support for other `java.lang.X` directly from JNI would be nice to have. For example `JInteger` -> `java.lang.Integer` that could have a `toDartInt` method.

This would mean that we need to come up with a different naming scheme to differentiate primitives and objects.

HosseinYousefi:
** This PR does not include the interop mechanism to actually include the new included types when generating using JNIgen for an easier and more focused review process. **

Adding several types:

* Changed primitive types to `jint`, `jshort`, ... format.
* Added
  * JInt, JLong, ... Boxed types
  * JMap
  * JSet
  * JList
And their related extension methods.


jni/lib/src/jni.dart
```diff
@@ -183,17 +185,13 @@ abstract class Jni {
...
-
-  /// Points to a process-wide shared instance of [GlobalJniEnvStruct].
+  /// Points to a process-wide shared instance of [GlobalJniEnv].
   ///
   /// It provides an indirection over [JniEnv] so that it can be used from
   /// any thread, and always returns global object references.
-  static GlobalJniEnv get env {
-    return _env ??= GlobalJniEnv(_fetchGlobalEnv());
-  }
+  static final env = GlobalJniEnv(_fetchGlobalEnv());
```
mahesh-hegde, line 192: Nice!

jni/lib/src/jobject.dart
```diff
@@ -2,7 +2,22 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-part of 'types.dart';
```
mahesh-hegde, line 5: Nice!

jni/lib/src/jreference.dart
```diff
@@ -2,44 +2,52 @@
...
-part of 'types.dart';
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+import 'package:jni/src/third_party/generated_bindings.dart';
+
+import 'jexceptions.dart';
+import 'jni.dart';
+
+extension ProtectedJReference on JReference {
```
mahesh-hegde, line 13: Turns out java was right all along!

jni/lib/src/lang/jboolean.dart
```diff
@@ -0,0 +1,89 @@
...
+  static JBoolean get falseValue => const JBooleanType().fromRef(Jni.accessors
+      .getStaticField(_class.reference, _falseId, JniCallType.objectType)
+      .object);
+
+  static final _trueId = Jni.accessors.getStaticFieldIDOf(
+    _class.reference,
+    r"TRUE",
+    r"Ljava/lang/Boolean;",
+  );
+  static JBoolean get trueValue => const JBooleanType().fromRef(Jni.accessors
```
mahesh-hegde, line 74: was this generated using jnigen? (Looking at `deleteOriginal` below I think not.)

I would be interested in an extension mechansim which lets us generate the class as usual, while adding our own extension methods in other files, then export our methods in from respective generated classes.

jni/lib/src/lang/jinteger.dart
```diff
@@ -0,0 +1,57 @@
...
+
+  @override
+  final superCount = 2;
+
+  @override
+  int get hashCode => (JIntegerType).hashCode;
+
+  @override
+  bool operator ==(Object other) {
+    return other.runtimeType == JIntegerType && other is JIntegerType;
```
mahesh-hegde, line 34: Doesn't `other.runtimeType == JIntegerType` imply `other is JIntegerType`?

jni/lib/src/lang/lang.dart
mahesh-hegde, line 1: There are certainly types I have seen more than `short` in the wild.

`CharSequence` (Android APIs, esp), `Class<T>` and `Iterable` from `java.lang`, but also some `java.io` types like `OutputStream` in some libraries, and of course, `Stream<>` and `Optional<>` in modern libraries.

Let's see how this plays out in practice, whether adding builtin types scales well or we have to resort to a semi-generated approach.

jni/lib/src/util/jlist.dart
```diff
@@ -0,0 +1,294 @@
...
+      .callMethodWithArgs(reference, _sizeId, JniCallType.intType, []).integer;
+
+  @override
+  set length(int newLength) {
+    RangeError.checkNotNegative(newLength);
+    while (length < newLength) {
+      add(E.fromRef(nullptr));
+    }
+    while (newLength < length) {
+      removeAt(length - 1);
```
mahesh-hegde, line 90: A high level question is whether you want Java list to behave like dart list here, or as java `List`.

Here we are not hiding implementation detail of JNI but interface of java itself.

@dcharkes wdyt?

jni/test/jmap_test.dart
```diff
@@ -0,0 +1,158 @@
...
+import 'package:test/test.dart';
+
+import 'test_util/test_util.dart';
+
+void main() {
+  // Don't forget to initialize JNI.
+  if (!Platform.isAndroid) {
+    try {
+      Jni.spawn(dylibDir: "build/jni_libs", jvmOptions: ["-Xmx128m"]);
+    } on JvmExistsException catch (_) {
```
mahesh-hegde, line 17: use `spawnIfNotExists`?

mahesh-hegde (approved):
Nice!

jni/lib/src/lang/jinteger.dart
HosseinYousefi: other is Something could mean that other extends Something as well.

jni/test/jmap_test.dart
HosseinYousefi: This is just a duplicate of existing tests. We can change them everywhere.

jni/lib/src/lang/jinteger.dart
mahesh-hegde: Then runtimeType will be most concrete subtype right?

HosseinYousefi: Yes. But you can override runtimeType to be anything. So in reality there is no sure fire way to check if something is of an exact type but this is the closest we can get.

mahesh-hegde: I see that's very pythonic.

jni/lib/src/jobject.dart
```diff
@@ -2,7 +2,22 @@
...
+// This typedef is needed because void is a keyword and cannot be used in
+// type switch like a regular type.
+import 'dart:ffi';
```
dcharkes, line 7: This is an import, not a typedef.

jni/lib/src/jprimitives.dart
```diff
@@ -2,117 +2,122 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-part of 'types.dart';
+// The types here are mapped to primitive types in Java, so they're all in
```
dcharkes, line 5: If this is documentation for the whole file, consider using

```dart
/// bla
library;
```

jni/lib/src/jreference.dart
dcharkes: ROFL!

jni/test/boxed_test.dart
```diff
@@ -0,0 +1,79 @@
...
+    const eps = 1e-9;
+    using((arena) {
+      expect(JDouble(val).doubleValue(deleteOriginal: true), closeTo(val, eps));
+      expect((-val).toJDouble().doubleValue(deleteOriginal: true),
+          closeTo(-val, eps));
+    });
+  });
+  testRunner('JBoolean', () {
+    using((arena) {
+      expect(JBoolean(true), JBoolean.trueValue);
```
dcharkes, line 73: `JBoolean(true)` is shorter than `JBoolean.trueValue`, why do we have the latter?

jni/test/jlist_test.dart
```diff
@@ -0,0 +1,197 @@
...
+      Jni.spawn(dylibDir: "build/jni_libs", jvmOptions: ["-Xmx128m"]);
+    } on JvmExistsException catch (_) {
+      // TODO(#51): Support destroying and reinstantiating JVM.
+    }
+  }
+  run(testRunner: test);
+}
+
+void run({required TestRunnerCallback testRunner}) {
+  JList<JString> newJList(Arena arena) {
```
dcharkes, line 25: `newJList` on the call-sites doesn't indicate it's test data. I was mistakenly thinking it would just give an empty list as part of the jni API.

Maybe call it `testDataList` or something.

jni/test/jmap_test.dart
```diff
@@ -0,0 +1,158 @@
...
+      // TODO(#51): Support destroying and reinstantiating JVM.
+    }
+  }
+  run(testRunner: test);
+}
+
+void run({required TestRunnerCallback testRunner}) {
+  JMap<JString, JString> newJMap(Arena arena) {
```
dcharkes, line 25: `testData`

Maybe also consider using setUp and tearDown and a late field for `testData`, then you don't have to invoke it in every test manually.

jni/test/jset_test.dart
```diff
@@ -0,0 +1,194 @@
...
+      Jni.spawn(dylibDir: "build/jni_libs", jvmOptions: ["-Xmx128m"]);
+    } on JvmExistsException catch (_) {
+      // TODO(#51): Support destroying and reinstantiating JVM.
+    }
+  }
+  run(testRunner: test);
+}
+
+void run({required TestRunnerCallback testRunner}) {
+  JSet<JString> newJSet(Arena arena) {
```
dcharkes, line 25: `testsData`

jni/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: jni
 description: Library to access JNI from dart and flutter
-version: 0.4.0
+version: 0.5.0-dev
```
dcharkes, line 3: `0.5.0-dev.0` is more common in the Dart eco system I believe.

It enables one to do dev releases and then bump the version number if need be.

jnigen/example/in_app_java/lib/android_utils.dart
```diff
@@ -27,7 +27,7 @@ final ffi.Pointer<T> Function<T extends ffi.NativeType>(String sym) jniLookup =
 /// from: com.example.in_app_java.AndroidUtils
 class AndroidUtils extends jni.JObject {
   @override
-  late final jni.JObjType $type = type;
+  late final jni.JObjType<AndroidUtils> $type = type;
```
dcharkes, line 30: Nice!

jnigen/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+## 0.5.0-dev
+
```
dcharkes, line 2: * No JNIGen changes (yet), but keeping version in lockstep with `package:jni`.

jnigen/analysis_options.yaml
```diff
@@ -18,3 +18,4 @@ linter:
     - prefer_const_constructors
     - prefer_relative_imports
 
+
```
dcharkes, line 21: spurious newlines

dcharkes (commented):
This PR could use some high level description:

* Were the files generated and then modified by hand? How will we maintain those?
* How will the follow up PR build on this?

HosseinYousefi:
> This PR could use some high level description:
> 
> * Were the files generated and then modified by hand? How will we maintain those?
> * How will the follow up PR build on this?

The files were manually written. I copied the method returns from the generated code.

There are not many classes added, and they're not following the usual Java style but rather they conform to Dart's interfaces, so not much maintenance is needed going forward. 

The following PR is going to be about a general interop mechanism between JNI packages. JNIgen generated code can implicitly interop with the existing code from package:jni. Similar to how currently we generate `jni.JString` when encountering `java.lang.String`. The classes added in this PR will be used similar to the Java's String class.

dcharkes:
> There are not many classes added, and they're not following the usual Java style but rather they conform to Dart's interfaces, so not much maintenance is needed going forward.

Famous last words! Just kidding, I think it's fine.

Just to be sure, if you made any copy errors we end up getting JVM errors I suppose?

I think 100% coverage is a must if it's not auto-generated, just to ensure we don't have any copy paste errors or typos.

Just a random check: https://coveralls.io/builds/59585492/source?filename=jni%2Flib%2Fsrc%2Flang%2Fjlong.dart#L16 we're missing fromRef, hashCode, ==, etc.

jni/lib/src/lang/jboolean.dart
HosseinYousefi: It's a combination. The type classes were directly copied. The method returns were copied but the classes and method definitions (implementing Dart abstract classes) were manually written and changed.

HosseinYousefi: cc @dcharkes

jni/test/boxed_test.dart
HosseinYousefi: Right, we don't need to have it.

HosseinYousefi:
> I think 100% coverage is a must if it's not auto-generated, just to ensure we don't have any copy paste errors or typos.

I can of course add tests to increase the coverage, however most of the uncovered methods are from `JXType` classes which were directly copied from the generated code. All the actual methods of the newly added `JX` classes are covered.

jni/lib/src/jprimitives.dart
HosseinYousefi: Requires to update the SDK to 2.19 to use unnamed libraries. I'll do it when we actually want to use other Dart 3 features in the future.

jni/lib/src/lang/lang.dart
HosseinYousefi: Let's see how it plays out! I'm open to add a few more types as long as it matches well with a Dart type, because these types are hard to "generate", they need to be manually written to give the same Dart feel.

jni/lib/src/util/jlist.dart
HosseinYousefi: We had a prior discussion about this and we decided to make them behave more like Dart lists. One of the reasonings being that you can always generate your own `java.lang.List` which is more Java like using `jnigen`.

jni/lib/src/lang/jboolean.dart
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: overridden_fields
```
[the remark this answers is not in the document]
dcharkes: Is the lint too strict, e.g. with final fields we're only overriding the getter, which should be perfectly fine? https://dart-lang.github.io/linter/lints/overridden_fields.html seems to only talk about non-final fields.

https://github.com/dart-lang/linter/issues/1109 says it's bad practise, because you could have side effects.

Maybe just ignore the lint for the line instead of the whole file, to document that we're intentionally overriding final fields with final fields and have no side effects.

dcharkes (approved):

dcharkes:
> > I think 100% coverage is a must if it's not auto-generated, just to ensure we don't have any copy paste errors or typos.
> 
> I can of course add tests to increase the coverage, however most of the uncovered methods are from `JXType` classes which were directly copied from the generated code. All the actual methods of the newly added `JX` classes are covered.

You are much more confident in your copy paste skills than I am in mine. 😆 🍻

HosseinYousefi:
> You are much more confident in your copy paste skills than I am in mine. 😆 🍻

I'll add the tests don't worry 😆

PR #272, opened 2023-05-07
Summarizer fix class listing

mahesh-hegde:
Summarizer was not listing packages recursively in ~some~ many cases.

Edit: Some more testing would be nice though.

Since it seems we are gonna need many tests, should move all `generates summary` tests to JUnit.

But characteristic tests should stay in dart, because deserialization still happens in dart.


jnigen/test/summary_generation_test.dart
HosseinYousefi, line 1: Could you also add some unhappy paths to the tests here? For example, if the class doesn't exists.

mahesh-hegde: I am wondering if it's better to do this in JUnit than in dart, which kinda slows down the tests.

But by testing it in junit, we wont be testing deserialization, and will have to rewrite the machinery. Since maven build doesn't have simple access to our test java sources, that will also be complicated.

HosseinYousefi: Sure.

mahesh-hegde: I actually added tests in dart itself.

HosseinYousefi (approved):
LGTM 🚀 Thanks @mahesh-hegde!

PR #279, opened 2023-05-13
Fix encoding bug

### LINKED ISSUE #278: UTF-8 strings not properly decoded
Passing UTF-8 strings to/from JNI doesn't seem to work as expected, although I could be misunderstanding things.

```dart
import 'dart:io';

import 'package:jni/jni.dart';

Future<void> main() async {
  Jni.spawn(
    dylibDir: Platform.script.resolve('../build/jni_libs').toFilePath(),
  );
  final utf8String = '🐬'.toJString();
  final dartString = utf8String.toDartString(deleteOriginal: true);
  print(dartString); // Prints ð
}
```

Does this have to do with the fact that JNI uses [Modified UTF-8](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/types.html#modified_utf_8_strings)?

Thanks for any help.

mahesh-hegde:
Use UTF-16 instead of UTF-8 in string methods.

closes #278

~GetStringCritical works on standalone but breaks on Android. I am still debugging the cause.~

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/jni/example/integration_test/on_device_jni_test.dart
+++ b/jni/example/integration_test/on_device_jni_test.dart
@@ -9,6 +9,7 @@ import '../../test/exception_test.dart' as exception_test;
 import '../../test/jlist_test.dart' as jlist_test;
 import '../../test/jmap_test.dart' as jmap_test;
 import '../../test/jobject_test.dart' as jobject_test;
+import '../../test/jstring_test.dart' as jstring_test;
 import '../../test/jset_test.dart' as jset_test;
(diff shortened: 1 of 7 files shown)


mahesh-hegde:
It's ready. I ended up using `GetStringChars` for the purpose of simplicity. When we have benchmarks we can try getStringCritical and see if there's an improvement.

HosseinYousefi (approved):
LGTM 🚀  Thanks, @mahesh-hegde!

PR #281, opened 2023-05-15
Fix the integration tests.

HosseinYousefi:
closes dart-lang/native#625

@mahesh-hegde  It's interesting that the normal tests didn't fail with a wrong return type for `add` method.

### DIFF
--- a/jni/example/android/build.gradle
+++ b/jni/example/android/build.gradle
@@ -26,6 +26,6 @@ subprojects {
     project.evaluationDependsOn(':app')
 }
 
-task clean(type: Delete) {
+tasks.register("clean", Delete) {
     delete rootProject.buildDir
 }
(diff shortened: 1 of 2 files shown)


dcharkes (approved):
gradle files 😨  ... run! 🏃‍♂️

HosseinYousefi:
Haha, same problem as https://github.com/dart-lang/native/pull/40#issuecomment-1547775663

Just gotta yoink the `apt-get update` here as well...

HosseinYousefi:
It already had the apt-get updates so just re-ran the tests...

mahesh-hegde:
> @mahesh-hegde It's interesting that the normal tests didn't fail with a wrong return type for add method.

A lot more spec violations fail on Android but pass silently on standalone. Part because of CheckJNI and some could be because default implementation is a little permissive too.

PR #289, opened 2023-05-31
Interop part 1

HosseinYousefi:
I decided to make the PR in multiple parts for an easier review. This part only includes the importing (and not generating the `symbols.yaml` file). And it only imports `JString` and `JObject` and none of the newly added `JList`, `JSet`, etc. This way none of the generated bindings have been changed in this PR.

### DIFF
--- a/jnigen/lib/src/bindings/dart_generator.dart
+++ b/jnigen/lib/src/bindings/dart_generator.dart
@@ -333,7 +333,7 @@ class _ClassGenerator extends Visitor<ClassDecl, void> {
         .map((typeParam) => 'this.$typeParam,')
         .join(_newLine(depth: 2));
     final superClass = (node.classDecl.superclass!.type as DeclaredType);
-    final superTypeClassesCall = superClass.classDecl.superCount == 0
+    final superTypeClassesCall = superClass.classDecl.isObject()
         ? ''
         : superClass.params
             .accept(_TypeClassGenerator(resolver))
@@ -1331,7 +1331,7 @@ class _TypeVarLocator extends TypeVisitor<Map<String, List<OutsideInBuffer>>> {
 
   @override
   Map<String, List<OutsideInBuffer>> visitDeclaredType(DeclaredType node) {
-    if (node.classDecl.superCount == 0) {
+    if (node.classDecl.isObject()) {
       return {};
     }
     final offset = node.classDecl.allTypeParams.length - node.params.length;
--- a/jnigen/lib/src/bindings/dart_generator.dart
+++ b/jnigen/lib/src/bindings/dart_generator.dart
@@ -31,9 +31,6 @@ const _voidPointer = '$_ffi.Pointer<$_ffi.Void>';
 
 // Prefixes and suffixes
 const _typeParamPrefix = '\$';
-// TODO(#143): this is a temporary fix for the name collision.
-const _typeClassPrefix = '\$';
-const _typeClassSuffix = 'Type';
 
 // Misc.
 const _classRef = '_class.reference';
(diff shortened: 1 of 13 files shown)


jnigen/lib/src/bindings/linker.dart
```diff
@@ -2,47 +2,95 @@
...
+    for (final className in config.importedClasses.keys) {
+      log.finest('Imported $className successfully.');
+    }
+
+    ClassDecl resolve(String? binaryName) {
+      return config.importedClasses[binaryName] ??
           node.decls[binaryName] ??
-          ClassDecl.object;
-    });
+          resolve('java.lang.Object');
```
dcharkes, line 62: Should we use something like `(TypeUsage.object.type as DeclaredType).binaryName` instead of repeating the string?
Should we have some special representation for Object? E.g. we default to it if we don't have a type. `ClassDecl.object`, `DeclaredType.object` etc.

Now that we have imports, maybe we shouldn't instantiate it manually (duplicating things between Dart/yaml), but just have a late final that does a `resolve` or something.

jnigen/test/package_resolver_test.dart
```diff
@@ -17,21 +18,28 @@ class ResolverTest {
 void main() async {
   await checkLocallyBuiltDependencies();
   final resolver = Resolver(
-      importMap: {
-        'org.apache.pdfbox': 'package:pdfbox/pdfbox.dart',
```
dcharkes, line 21: Since this is a breaking change, please bump the version. (It feels a bit ugly to keep supporting the import map, so probably best to do a breaking change.)

I see we haven't released a 0.5 yet, but there are fixes between 0.4 and 0.5, please make a PR to release 0.5 with the fixes in the changelog and then bump to 0.6.0-dev.0 in this PR.

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -573,19 +575,13 @@ class _TypeClassGenerator extends TypeVisitor<_TypeClass> {
...
     return _TypeClass(
-      '$ifConst$_jArray$_typeClassSuffix(${innerTypeClass.name})',
+      '$ifConst${_jArray}Type(${innerTypeClass.name})',
       innerTypeClass.canBeConst,
     );
   }
 
   @override
   _TypeClass visitDeclaredType(DeclaredType node) {
-    if (node.classDecl.binaryName == 'java.lang.Object' ||
```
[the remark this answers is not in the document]
HosseinYousefi: Yes, both `java.lang.String` and `java.lang.Object` are now a part of imported classes instead.

jnigen/lib/src/bindings/linker.dart
HosseinYousefi: As you can see in the red, I had the `ClassDecl.object` before. I think it's misplaced as a static field if we can realistically change it, maybe `config.object` is better. If we choose to never change `object`, then we can keep it hardcoded.

dcharkes: Okay, I leave it to you to find the cleanest solution. 👌

jnigen/lib/src/bindings/linker.dart
```diff
@@ -2,47 +2,95 @@
...
+class Linker extends Visitor<Classes, Future<void>> {
   Linker(this.config);
 
   final Config config;
 
   @override
-  void visit(Classes node) {
-    final classLinker = _ClassLinker(config, (binaryName) {
-      return ClassDecl.predefined[binaryName] ??
+  Future<void> visit(Classes node) async {
```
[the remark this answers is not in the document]
dcharkes: Ack!

dcharkes (approved):
lgtm! Thanks @HosseinYousefi !

dcharkes (approved):
lgtm! Thanks @HosseinYousefi !

PR #297, opened 2023-06-07
Non static nested class construction

HosseinYousefi:
Take a non-static nested class like `B`:

```java
public class A {
  public class B {
    // ...
    public B(int a, int b) { /* ... */ }
  }
}
```

Right now, we're incorrectly generating `B`'s constructor like `B(int a, int b)`, but since `B` is not static, it depends on its outer class and has to take it as its first parameter. This PR fixes this issue so `B`'s constructor turns into `B(A $parent, int a, int b)` instead.

---

We could make the constructors for the nested class private, and generate a temporary class `$BBuilder` which takes an instance of `A` as it's constructor and use it as a `B` getter in `A`:

```dart
class A {
  // ...
  $BBuilder get B => $BBuilder(this);
}

class $BBuilder {
  final A $parent;
  $BBuilder(this.$parent);

  B call(int a, int b) {
    return B._($parent, a, b);
  }
}
```

Although making the syntax of creating the nested class nicer (`final b = a.B(2, 3);` instead of `final b = B(a, 2, 3);`), it makes the code generation more complicated and it's not the highest priority. So this feature is not included in this PR.

### DIFF
--- a/jnigen/lib/src/bindings/unnester.dart
+++ b/jnigen/lib/src/bindings/unnester.dart
@@ -1,0 +1,90 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import '../elements/elements.dart';
+import 'visitor.dart';
+
+/// A [Visitor] that processes nested classes.
+///
+/// Nested classes are not supported in Dart. So this "unnests" them into
+/// separate classes.
+class Unnester extends Visitor<Classes, void> {
+  @override
+  void visit(node) {
+    final classProcessor = _ClassUnnester();
+    for (final classDecl in node.decls.values) {
+      classDecl.accept(classProcessor);
+    }
+  }
+}
+
+class _ClassUnnester extends Visitor<ClassDecl, void> {
+  final processed = <ClassDecl>{};
+
+  @override
+  void visit(ClassDecl node) {
+    if (processed.contains(node)) return;
+    processed.add(node);
+    node.parent?.accept(this);
+
+    // Add type params of outer classes to the nested classes
+    final allTypeParams = <TypeParam>[];
+    if (!node.isStatic) {
+      for (final typeParam in node.parent?.allTypeParams ?? []) {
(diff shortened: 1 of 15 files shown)


jnigen/test/simple_package_test/dart_only/dart_bindings/simple_package.dart
```diff
@@ -2084,27 +2091,32 @@ class MyMap_MyEntry<$K extends jni.JObject, $V extends jni.JObject>
...
-      _class.reference, r"<init>", r"(Ljava/lang/Object;Ljava/lang/Object;)V");
+      _class.reference,
+      r"<init>",
+      r"(Lcom/github/dart_lang/jnigen/generics/MyMap;Ljava/lang/Object;Ljava/lang/Object;)V");
 
-  /// from: public void <init>(K key, V value)
+  /// from: public void <init>(com.github.dart_lang.jnigen.generics.MyMap $parent, K key, V value)
   /// The returned object must be deleted after use, by calling the `delete` method.
   factory MyMap_MyEntry(
+    MyMap<$K, $V> $parent,
```
dcharkes, line 2101: How was this correct before? :)

dcharkes (approved):
lgtm with comments!

jnigen/test/simple_package_test/dart_only/dart_bindings/simple_package.dart
HosseinYousefi: It was not 😛

jnigen/lib/src/bindings/unnester.dart
```diff
@@ -0,0 +1,90 @@
...
+    if (processed.contains(node)) return;
+    processed.add(node);
+    node.parent?.accept(this);
+
+    // Add type params of outer classes to the nested classes
+    final allTypeParams = <TypeParam>[];
+    if (!node.isStatic) {
+      for (final typeParam in node.parent?.allTypeParams ?? []) {
+        if (!node.allTypeParams.map((t) => t.name).contains(typeParam.name)) {
+          // Add only if it's not shadowing another type param.
```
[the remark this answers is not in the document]
HosseinYousefi: Added a TODO to a bug, because in case of shadowing the type inference also stops working.

PR #326, opened 2023-07-19
Interface implementation

HosseinYousefi:
The `PortProxy` is a proxy class constructed with a native port, a function pointer, and a thread id. When an interface method is invoked, it runs the `_invoke` native method. If the thread id of the invocation is the same as the construction thread id, the function pointer will be called directly. Otherwise we use the send port + a condition variable to synchronize and get the result back.

Static methods are quite clunky to use in real world applications, so developers can pass in normal callbacks to `.implement` method of the interface. For the `ReceivePort` mode, this is straight-forward. However for the direct function pointer call, we keep the callbacks and type classes in `_$methods` and `_$types` static maps.

TODO: more tests like
- [x] Test the other two functions in the interface
- [x] Calling the interface method through Dart directly
- [x] Recursive
- [x] Mixing same thread calls + different thread calls

~- [ ] Add an example that uses it~ Will add once the API is stable.


jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -528,6 +529,59 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
+        myInterface,
+        'hello'.toJString(),
+      );
+      final result = await voidCallbackResult.future;
+      expect(result.toDartString(deleteOriginal: true), 'hellohello');
+      myInterface.delete();
+    });
+    test('MyInterface.implement consume on the same thread', () async {
+      final voidCallbackResult = Completer<JString>();
+      final myInterface = MyInterface.implement(
```
dcharkes, line 561: I believe we should try to make it an actual interface implementation rather than passing closures for everything.

```dart
abstract class MyInterface {
  void Function voidCallback(jni.JString s);
  jni.JString stringCallback(jni.JString s);
  // ..
}
```

And then you'd implement it:

```dart
class MyInterfaceImplementation extends MyInterface {
  final voidCallbackResult = Completer<JString>();

  void Function voidCallback(jni.JString s){
    voidCallbackResult.complete();
  }

  // ...
}
```

The implementation would then be instead of holding on to `_$methods` hold on to an object and use tearoffs rather than closures.

I believe this would be a better API and better performance.
Is there something I've missed about why we do it closure based?

dcharkes (changes requested):
I believe we should try to make it an actual interface implementation rather than passing closures for everything.

```dart
abstract class MyInterface {
  void Function voidCallback(jni.JString s);
  jni.JString stringCallback(jni.JString s);
  // ..
}
```

And then you'd implement it:

```dart
class MyInterfaceImplementation extends MyInterface {
  final voidCallbackResult = Completer<JString>();

  void Function voidCallback(jni.JString s){
    voidCallbackResult.complete();
  }

  // ...
}
```

The implementation would then be instead of holding on to `_$methods` hold on to an object and use tearoffs rather than closures.

Maybe we should actually try to use implements even:

```dart
class MyInterfaceImplementation extends JObject implements MyInterface {
```

I believe this would be a better API and better performance.
Is there something I've missed about why we do it closure based?

HosseinYousefi:
> I believe this would be a better API and better performance. Is there something I've missed about why we do it closure based?

Let's say we extend/implement a class, sure now we keep the functions in an object instead of a map. But how can the user construct this class? They would have to write their own constructor, and write their own machinery. At least until we can do https://github.com/dart-lang/language/issues/1803.

jnigen/test/kotlin_test/generate.dart
```diff
@@ -63,7 +63,7 @@ Config getConfig([BindingsType bindingsType = BindingsType.cBased]) {
         structure: OutputStructure.singleFile,
       ),
     ),
-    summarizerOptions: SummarizerOptions(backend: 'asm'),
+    summarizerOptions: SummarizerOptions(backend: SummarizerBackend.asm),
```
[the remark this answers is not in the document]
HosseinYousefi: I actually now removed the only `jni.jar` we had. Now I'm auto generating it using `dart run jni:setup` in the `build/` folder which is ignored.

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -399,6 +401,117 @@ class $name$typeParamsDef extends $superName {
...
     }
 
+    // Interface implementation
+    if (node.declKind == DeclKind.interfaceKind) {
+      s.write('''
+  /// Maps a specific port to the implemented methods.
+  static final Map<int, Map<String, Function>> _\$methods = {};
+
+  /// Maps a specific port to the type parameters.
+  static final Map<int, Map<String, $_jType>> _\$types = {};
```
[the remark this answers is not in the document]
HosseinYousefi: I can improve the generated code, it's somewhat of a rule to always use normal `'''` throughout the binding generation code, I can easily add interpolations this way.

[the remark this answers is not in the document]
HosseinYousefi: I thought of using mixins too. I'll explore this further after the initial discussions.

dcharkes:
> > I believe this would be a better API and better performance. Is there something I've missed about why we do it closure based?
> 
> Let's say we extend/implement a class, sure now we keep the functions in an object instead of a map. But how can the user construct this class? They would have to write their own constructor, and write their own machinery. At least until we can do [dart-lang/language#1803](https://github.com/dart-lang/language/issues/1803).

Can we provide a mixin or a base class for that machinery?

jni/android/src/main/java/com/github/dart_lang/jni/PortProxy.java
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
HosseinYousefi: The reason we had two versions was to automatically have it in Flutter Android + the `@Keep` annotation (which is in `androidx`) so proguard doesn't remove it. I learned enough gradle to remove the duplication and add proguard rules in a different way!

HosseinYousefi:
Put the interface implementation related things under an experiment flag, so we can continue testing it in our CI while not exposing the API until it's ready.

README.md
```diff
@@ -218,6 +218,7 @@ A `*` denotes required configuration.
 | `source_path`          | List of directory paths | Directories to search for source files. Note: source_path for dependencies downloaded using `maven_downloads` configuration is added automatically without the need to specify here. |
 | `class_path`           | List of directory / JAR paths | Classpath for API summary generation. This should include any JAR dependencies of the source files in `source_path`. |
 | `classes` *            | List of qualified class / package names | List of qualified class / package names. `source_path` will be scanned assuming the sources follow standard java-ish hierarchy. That is a.b.c either maps to a directory `a/b/c` or a class file `a/b/c.java`.  |
+| `enable_experiment`    | List of experiment names | List of enabled experiments. These features are still in development and their API might break.  |
```
dcharkes, line 221: Should this table list the available experiments?

jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -528,6 +529,74 @@ void registerTests(String groupName, TestRunnerCallback test) {
       });
     });
   });
+
+  group('interface implementation', () {
+    for (final method in {
+      'another thread': MyInterfaceConsumer.consumeOnAnotherThread,
```
dcharkes, line 535: It would be nicer to use records instead of a map here, it's easier to read on the callsite. However, that means forcing the SDK lower bound to 3.0.

jnigen/lib/src/config/config_types.dart
```diff
@@ -619,6 +652,19 @@ class Config {
   }
 }
 
+enum Experiment { interfaceImplementation }
```
dcharkes, line 655: Can we make this a class with two fields `configString` and `description`?

For reference (we probably don't need something as big as these two).

Dart experiments:

```dart
  nativeAssets(
      name: 'native-assets',
      isEnabledByDefault: false,
      isExpired: false,
      experimentEnabledVersion: const Version(3, 1),
      experimentReleasedVersion: const Version(3, 1)),
```

Flutter experiments:

```dart
const Feature nativeAssets = Feature(
  name: 'native assets compilation and bundling',
  configSetting: 'enable-native-assets',
  environmentOverride: 'FLUTTER_NATIVE_ASSETS',
  master: FeatureChannelSetting(
    available: true,
  ),
);
```

jni/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.6.0-dev-2
```
dcharkes, line 1: I recently learned we should use `-dev` for when making an actual release, and `-wip` for landing commits that we don't make a release for yet.

https://github.com/dart-lang/native/pull/89#issuecomment-1633204927

I believe this should be `-wip` until we remove the `-wip` and release `0.6.0`. We're not intending to release a `0.6.0-dev-2` before `0.6.0`.

dcharkes (approved):

jnigen/test/simple_package_test/runtime_test_registrant.dart
HosseinYousefi: I'll make a slightly bigger change to the codebase after upgrading to 3.0.0 in a follow-up PR.

PR #330, opened 2023-07-25
Fix CI

HosseinYousefi:


jnigen/test/test_util/test_util.dart
```diff
@@ -88,9 +88,10 @@ void comparePaths(String path1, String path2) {
   if (diffProc.exitCode != 0) {
     final originalDiff = diffProc.stdout;
     log.warning(
-        "Paths $path1 and $path2 differ, comparing by ignoring space change");
-    final fallbackDiffProc = Process.runSync(
-        "git", [...diffCommand, '--ignore-space-change', path1, path2]);
+        "Paths $path1 and $path2 differ, Running dart format on $path1.");
+    Process.runSync('dart', ['format', path1]);
+    final fallbackDiffProc =
+        Process.runSync("git", [...diffCommand, path1, path2]);
```
dcharkes, line 94: If we run the formatter, shouldn't we just be able to a literal string compare instead?

Unless you want to actually show the diff after formatting, but you're showing the originalDiff two lines below.

dcharkes (approved):

jnigen/test/test_util/test_util.dart
dcharkes: If we format both files, we don't have to git diff, we can just load the two files as strings and check equality.

HosseinYousefi: There can be multiple files in a path.

jnigen/test/test_util/test_util.dart
```diff
@@ -88,9 +88,10 @@ void comparePaths(String path1, String path2) {
...
     if (fallbackDiffProc.exitCode != 0) {
       stderr.writeln(originalDiff);
```
dcharkes, line 96: Wouldn't you want to print `fallbackDiffProc.stdout`/`fallbackDiffProc.stderr` here?

HosseinYousefi: `dart format` is an exceptional thing that happens only when stable != beta. When even after formatting the diff produces a non-zero result, I show whatever happened originally before formatting.

jnigen/test/test_util/test_util.dart
dcharkes: Ah, check.

jnigen/test/test_util/test_util.dart
dcharkes: Shouldn't the fallback diff be more useful? Otherwise you get all the things that are not actually diffs.

HosseinYousefi: I understand your point. My logic is: if beta fails, stable will surely fail too, so we get the good diff there. `dart format` is only a hack to fix the CI in our current situation. And if this is no longer the case, we're falling back to the original way. (As if we never had the `dart format`).

PR #335, opened 2023-07-29
Fix windows

### LINKED ISSUE #332: Fix interface implementation test on Windows

HosseinYousefi:
Closes #332.

Also fixes the leaking of global references by cleaning them up after use in Java.

---

The fix for windows is a single line removal from dartjni.h. Windows doesn't have a function to destroy condition variable but I accidentally used the function to destroy the mutex for it. See the commit with 🤦‍♂️!

--- 

The rest of the PR is fixing a memory ownership problem. Instead of attaching the return value of the implemented method to a NativeFinalizer. We are passing the global reference to Java first and get an object from it. And then deleting the global reference used.

We can't do this cleanly in a single native function because if we delete the global reference, we can no longer return the result object.

Since Dart doesn't know that this global reference is still used, it might garbage collect it via NativeFinalizer thus making it invalid. So I'm letting Java handle the clean up and use setAsDeleted to detach the native finalizer from the return value in Dart.


jni/java/src/main/java/com/github/dart_lang/jni/PortProxy.java
```diff
@@ -62,21 +62,28 @@ private static void appendType(StringBuilder descriptor, Class<?> type) {
...
     return Proxy.newProxyInstance(
         clazz.getClassLoader(), new Class[] {clazz}, new PortProxy(port, threadId, functionPtr));
   }
 
   @Override
-  public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
-    return _invoke(port, threadId, functionPtr, proxy, getDescriptor(method), args);
+  public Object invoke(Object proxy, Method method, Object[] args) {
+    Object[] result = _invoke(port, threadId, functionPtr, proxy, getDescriptor(method), args);
+    _cleanUp((Long) result[0]);
```
dcharkes, line 73: Why does calling two native methods instead of 1 make a difference? Shouldn't we be able to do in a single C function what we can do in two?

jni/src/dartjni.c
```diff
@@ -674,18 +678,36 @@ Java_com_github_dart_1lang_jni_PortProxy__1invoke(JNIEnv* env,
...
         port, (*env)->NewGlobalRef(env, methodDescriptor),
         (*env)->NewGlobalRef(env, args));
   }
-}
\ No newline at end of file
+  attach_thread();
+  load_class_global_ref(&_c_Object, "java/lang/Object");
+  load_class_global_ref(&_c_Long, "java/lang/Long");
+  load_method(_c_Long, &_m_Long_init, "<init>", "(J)V");
+  jobject first = (*env)->NewObject(env, _c_Long, _m_Long_init, (jlong)result);
```
dcharkes, line 698: If I understand correctly, we're boxing here. Why do we need boxing for Windows?

dcharkes (commented):
There's no description on the PR what we're fixing (and I have trouble reverse engineering the issue from the diff 🙃).

HosseinYousefi:
> There's no description on the PR what we're fixing (and I have trouble reverse engineering the issue from the diff 🙃).

The fix for windows is a single line removal from dartjni.h. Windows doesn't have a function to destroy condition variable but I accidentally used the function to destroy the mutex for it. See the commit with 🤦‍♂️!

---

The rest of the PR is fixing a memory ownership problem. Instead of attaching the return value of the implemented method to a NativeFinalizer. We are passing the global reference to Java first and get an object from it. And then deleting the global reference used. 

We can't do this cleanly in a single native function because if we delete the global reference, we can no longer return the result object. 

Since Dart doesn't know that this global reference is still used, it might garbage collect it via NativeFinalizer thus making it invalid. So I'm letting Java handle the clean up and use setAsDeleted to detach the native finalizer from the return value in Dart.

dcharkes:
> The rest of the PR is fixing a memory ownership problem. Instead of attaching the return value of the implemented method to a NativeFinalizer. We are passing the global reference to Java first and get an object from it. And then deleting the global reference used.
> 
> We can't do this cleanly in a single native function because if we delete the global reference, we can no longer return the result object.
> 
> Since Dart doesn't know that this global reference is still used, it might garbage collect it via NativeFinalizer thus making it invalid. So I'm letting Java handle the clean up and use setAsDeleted to detach the native finalizer from the return value in Dart.

Ah, that makes sense. 👍  Please add that as some documentation in the code and on the PR.

dcharkes (approved):

jni/src/dartjni.c
HosseinYousefi: We're returning an array of Objects so unfortunately, we have to box the long value.

jni/java/src/main/java/com/github/dart_lang/jni/PortProxy.java
HosseinYousefi: Explained below.

PR #349, opened 2023-08-11
Clean up interface implementation using auxiliary generated classes

HosseinYousefi:
Creating two extra classes for each interface when the `interface_implementation` experiment is enabled:

1. `$<Interface>Impl`: This abstract (interface can be added after upgrade to Dart 3 – #334) class can be implemented and then used as an argument for `<Interface>.implement`, instead of passing closures one by one.
2. `_$<Interface>Impl`: This class is a concrete implementation of the previous which gets individual closures like before. This is private and its constructor is exposed through the unnamed factory for `$<Interface>Impl`. This is useful for creating a quick lambda-like class.

Although we still have not explored the direct extension of the Java object, namings, etc. This PR makes the code cleaner that what it currently is and gives users the ability to neatly organize their code without sacrificing the convenience of creating inline classes.

I will create a design doc exploring the other parts of the design. We should also have a way to make `void` methods non-blocking. Something like `bool get $<method-name>Blocking;` that can be set to either `true` or `false`.

### DIFF
--- a/jnigen/lib/src/bindings/dart_generator.dart
+++ b/jnigen/lib/src/bindings/dart_generator.dart
@@ -15,11 +15,11 @@ import 'c_generator.dart';
 import 'resolver.dart';
 import 'visitor.dart';
 
-// Import prefixes
+// Import prefixes.
 const _jni = 'jni';
 const _ffi = 'ffi';
 
-// package:jni types
+// package:jni types.
 const _jType = '$_jni.JObjType';
 const _jPointer = '$_jni.JObjectPtr';
 const _jArray = '$_jni.JArray';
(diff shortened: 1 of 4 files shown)


jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -518,6 +498,83 @@ class $name$typeParamsDef extends $superName {
...
+  $typeClassGetters
+
+''');
+      final abstractImplMethod = _AbstractImplMethod(resolver, s);
+      for (final method in node.methods) {
+        method.accept(abstractImplMethod);
+      }
+      s.writeln('}');
+
+      // Concrete Impl class
```
dcharkes, line 536: nit: punctuation.

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -46,6 +46,7 @@ const _deleteInstruction =
     '  /// The returned object must be deleted after use, '
     'by calling the `delete` method.';
 
+// Helper methods
```
dcharkes, line 49: nit: The comment doesn't add anything, unnamed extensions are scoped to the file. Maybe just remove the comment.

dcharkes (approved):
Nice!

PR #350, opened 2023-08-13
feat: ability to add custom repos

gergovari:
Ability to add custom repositories to Maven, currently it's prefilled in the POM stub.

---

- [X] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/jnigen/bin/download_maven_jars.dart
+++ b/jnigen/bin/download_maven_jars.dart
@@ -14,10 +14,12 @@ void main(List<String> args) async {
   final config = Config.parseArgs(args);
   final mavenDownloads = config.mavenDownloads;
   if (mavenDownloads != null) {
+    stdout.writeln(mavenDownloads.repos.length);
+    stdout.writeAll(mavenDownloads.repos);
     await MavenTools.downloadMavenJars(
         MavenTools.deps(mavenDownloads.jarOnlyDeps + mavenDownloads.sourceDeps),
         mavenDownloads.jarDir,
-        repos: mavenDownloads.repos);
+        repos: MavenTools.repos(mavenDownloads.repos));
     await Directory(mavenDownloads.jarDir)
         .list()
         .map((entry) => entry.path)
--- a/jnigen/bin/download_maven_jars.dart
+++ b/jnigen/bin/download_maven_jars.dart
@@ -16,7 +16,8 @@ void main(List<String> args) async {
   if (mavenDownloads != null) {
     await MavenTools.downloadMavenJars(
         MavenTools.deps(mavenDownloads.jarOnlyDeps + mavenDownloads.sourceDeps),
-        mavenDownloads.jarDir);
+        mavenDownloads.jarDir,
+        repos: mavenDownloads.repos);
     await Directory(mavenDownloads.jarDir)
         .list()
         .map((entry) => entry.path)
--- a/jnigen/lib/src/config/config_types.dart
+++ b/jnigen/lib/src/config/config_types.dart
@@ -9,7 +9,6 @@ import 'package:path/path.dart' as p;
 import 'package:pub_semver/pub_semver.dart';
 import 'package:yaml/yaml.dart';
 
-import '../../tools.dart';
 import '../elements/elements.dart';
 import '../logging/logging.dart';
 import '../util/find_package.dart';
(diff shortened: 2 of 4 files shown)


gergovari:
I'm sure there is a lot of thinking left on the table towards the attitude of configs.

- I may get the idea for using a string for maven deps, but it was out of my laziness that I did this for repos. (Pretty ugly ngl): I would need to play around with YamlMaps.

jnigen/bin/download_maven_jars.dart
```diff
@@ -14,9 +14,12 @@ void main(List<String> args) async {
   final config = Config.parseArgs(args);
   final mavenDownloads = config.mavenDownloads;
   if (mavenDownloads != null) {
+    stdout.writeln(mavenDownloads.repos.length);
```
mahesh-hegde, line 17: You left some `print`s here. Remove them.

jnigen/lib/src/config/config_types.dart
```diff
@@ -34,17 +34,18 @@ class MavenDownloads {
...
-    this.jarOnlyDeps = const [],
-    this.jarDir = defaultMavenJarDir,
-  });
+  MavenDownloads(
+      {this.sourceDeps = const [],
+      // ASK: Should this be changed to a gitignore'd directory like build ?
+      this.sourceDir = defaultMavenSourceDir,
+      this.jarOnlyDeps = const [],
+      this.jarDir = defaultMavenJarDir,
+      this.repos = const []});
```
mahesh-hegde, line 43: You sure this is formatted correctly using `dart format` or `flutter format`?

jnigen/lib/src/tools/maven_tools.dart
```diff
@@ -126,17 +153,29 @@ ${depDecls.join("\n")}
...
     return MavenDependency(components[0], components[1], components[2]);
   }
   String groupID, artifactID, version;
   Map<String, String> otherTags;
 }
+
+class MavenRepository {
+  MavenRepository(this.id, this.name, this.url, {this.otherTags = const {}});
+  factory MavenRepository.fromString(String fullName) {
+    final components = fullName.split('|');
```
mahesh-hegde, line 173: This is an undocumented syntax to use `|` as separator?

In case of maven artifacts, the `:` separator between group, artifact and version is universally used convention.

I'd rather read this from YAML alone. (If there's no need to specify it from CLI).

```yaml
repos:
  - id: google
    name: Google Maven Repo
    url: https://maven.google.com
```

jnigen/lib/src/tools/maven_tools.dart
```diff
@@ -126,17 +153,29 @@ ${depDecls.join("\n")}
...
 }
 
-/// Maven dependency with group ID, artifact ID, and version.
 class MavenDependency {
   MavenDependency(this.groupID, this.artifactID, this.version,
       {this.otherTags = const {}});
   factory MavenDependency.fromString(String fullName) {
     final components = fullName.split(':');
-    if (components.length != 3) {
+    if (components.length < 3) {
```
mahesh-hegde, line 161: What's the idea behind this change?

I think If we don't handle all the optional appends in maven coordinate syntax, rather not allow them in the first place.

mahesh-hegde (changes requested):
Thanks for the contribution. You will require an approving review from HosseinYousefi or dcharkes.

Few notes:

1. If you make a change which adds a feature, or changes a design, its suggested to discuss with the maintainers through an issue. This is because getting good design is harder than writing code.

    In this case, we should've discussed things such as

    * Why this feature is required? What are the alternatives. (Eg: If you're using Maven from a company proxy, it might be better to set your company repo as maven global configuration).

    * What's the best way to expose this option to the user? In this PR, you used `|` as the separator, which may not be the best way.

    * How to implement this? Usually this point is straightforward. But with more complicated features we will also discuss this.

2. Ideally, accompany any new feature with tests. Otherwise, whatever you implemented can break when someone is making code changes.
    * In this case, you can add a happy path test case, of using a public popular repo, and ensure the JAR is downloaded. But also make sure to skip the test in case of network error.

3. Add the new configuration options to the documentation table in README.md, preferably with an example.

HosseinYousefi:
Thank you for your contribution, @gergovari! 

I don't think it's a good idea to add more and more maven features to be configured through JNIgen. You can create your own pom.xml file and configure it to your liking. A script that first copies the maven dependencies and then runs JNIgen could be created for an individual project.

The maven download features have been added to improve the experience of getting started. But even that might be removed in favor of some documentation in the future.

gergovari:
I've to agree, the custom pom sounds a lot better.

PR #351, opened 2023-08-13
feat: ability to specify dep type

gergovari:
Ability to specify dependency type in the POM.

I need to download AAR deps in a plugin, I've not yet figured out how to include them in the classpath though.
(And I don't want to specify deps at the app module level, I'd like this plugin to be plug-and-play.)

Edit: this does work if I direct `example/`'s gradle to `mvn_jar` as a repo and specify the AAR dependency in the app module, but this isn't the best.


HosseinYousefi:
Thank you for your contribution, @gergovari! The same idea of not really be willing to add more maven features to JNIgen (and therefore having to maintain them as well.) Feel free to open up issues discussing the implementation ideas before opening up a PR in the future.

PR #360, opened 2023-08-17
Use isolate_id instead of thread_id

HosseinYousefi:
To decide whether or not we should call the function pointer, or send a message.


jni/src/dartjni.c
```diff
@@ -623,7 +623,7 @@ JniResult PortProxy__newInstance(jobject binaryName,
     return (JniResult){.value = {.j = 0}, .exception = check_exception()};
   jobject _result = (*jniEnv)->CallStaticObjectMethod(
       jniEnv, _c_PortProxy, _m_PortProxy__newInstance, binaryName, port,
-      thread_id(), functionPtr);
+      (jlong)Dart_CurrentIsolate_DL(), functionPtr);
   return to_global_ref_result(_result);
```
dcharkes, line 627: Update the pubspec.yaml to require Dart 3.1 which has the dart_api_dl.h version 2.3.

dcharkes (approved):
lgtm with comments

Issue #359, opened 2023-08-17
`ensureNotDeleted` not called in the generated bindings


HosseinYousefi:
We're not calling `ensureNotDeleted` to throw `UseAfterFree`.

One idea to make sure any method call will have checked for this, is to change `JReference`'s `final reference` into the pair of `final _reference` and `Pointer<Void> get reference { ensureNotDeleted(); return _reference; }`

@mahesh-hegde @dcharkes Would this cause too many unnecessary checks?

dcharkes:
From an encapsulation point of view, that would be a nicer solution indeed. (With the only caveat that you shouldn't hang on to that reference in a var in a long block of code where you might delete in between multiple uses: so that should go in the doc comment on the getter.)

I wouldn't worry about performance. most likely the small method will get inlined and it's just a bool check.

PR #361, opened 2023-08-17
Check for frees

HosseinYousefi:
As discussed in #359, I removed `ensureNotDeleted` and instead included it in the `reference` getter. The previous tests testing for exceptions, will test this change.

Also changing the pubspec version to fix the beta CI.


jnigen/example/pdfbox_plugin/example/macos/Flutter/ephemeral/Flutter-Generated.xcconfig
```diff
@@ -0,0 +1,11 @@
+// This is a generated file; do not edit or check into version control.
```
dcharkes, line 1: Looks like we're missing a gititgnore somewhere.

jnigen/example/pdfbox_plugin/example/macos/Flutter/ephemeral/flutter_export_environment.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+# This is a generated file; do not edit or check into version control.
```
dcharkes, line 2: ditto

jni/lib/src/jexceptions.dart
```diff
@@ -15,21 +15,21 @@ class UseAfterFreeException implements JException {
...
   String toString() {
-    return "use after free on $ptr through $object";
+    return "Use after free on $ptr through $object.";
   }
 }
 
 class NullJStringException implements JException {
   @override
-  String toString() => 'toDartString called on null JString reference';
+  String toString() => 'toDartString was called on null JString reference.';
```
dcharkes, line 24: Maybe put `toDartString` in backticks or quotes.

dcharkes (approved):

jni/lib/src/jreference.dart
```diff
@@ -27,31 +23,41 @@ extension ProtectedJReference on JReference {
...
-    _finalizer.attach(this, reference, detach: this);
+  JReference.fromRef(this._reference) {
+    _finalizer.attach(this, _reference, detach: this);
   }
 
   bool _deleted = false;
 
   /// Check whether the underlying JNI reference is `null`.
-  bool get isNull => reference == nullptr;
+  bool get isNull => _reference == nullptr;
```
HosseinYousefi, line 36: Should this be `reference` or `_reference`?

HosseinYousefi: @dcharkes

jnigen/example/pdfbox_plugin/example/macos/Flutter/ephemeral/Flutter-Generated.xcconfig
HosseinYousefi: macos is not supported yet, so I'll just remove it. It was accidentally added after the pub version change + upgrade.

jni/lib/src/jreference.dart
HosseinYousefi: I guess it doesn't really matter to check it here as well, so I'll go with `_reference`.

dcharkes: Looking from the tests, the uses actually are null (for example in the list test). If we don't want these references to change value over time (also not by deleting), I'd say it should be `reference`.

PR #362, opened 2023-08-18
Change ctor to new

### LINKED ISSUE #354: Naming of methods and constructors
Java classes can have multiple (unnamed) constructors. Currently we name the other constructors as `ctor1`, `ctor2`, ... . I propose changing the names to `new1`, `new2`, ... .

Rationale: Currently if we have multiple Java methods with the same name like `foo`, we name them as `foo`, `foo1`, `foo2`, ... meaning we could write:

```dart
final foo = clazz.foo;
final foo1 = clazz.foo1;
// ...
```

Similarly we could 
```dart
final constructor = clazz.new;
// but currently
final constructor1 = clazz.ctor1;
// after this change
final constructor2 = clazz.new2; // similar to clazz.new
```
@dcharkes @mahesh-hegde @lrhn Any better idea for naming multiple constructors / methods instead of the added number? Do you like `ctor` or `new` better?

HosseinYousefi:
Closes #354. Decided to go with `new` and keeping the numeration for now.

### DIFF
--- a/jnigen/CHANGELOG.md
+++ b/jnigen/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 0.6.0-wip.3
+* **Breaking Change** ([#354](https://github.com/dart-lang/jnigen/issues/354)): Renamed constructors from `ctor1`, `ctor2`, ... to `new1`, `new2`, ...
+
 ## 0.6.0-wip.2
 * Fixed a bug where the nested classes would be generated incorrectly depending on the backend used for generation.
 * Fixed a bug where ASM backend would produce the incorrect parent for multi-level nested classes.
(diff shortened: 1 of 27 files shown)


jnigen/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 0.6.0-wip.3
+* **Breaking Change** ([#354](https://github.com/dart-lang/jnigen/issues/354)): Renamed constructors from `ctor1`, `ctor2`, ... to `new1`, `new2`, ...
```
dcharkes, line 2: In the generated files I see `new0`?

HosseinYousefi: `_new0` is the private name for unnamed constructor. It won't be visible to the users.

dcharkes (approved):
lgtm! Thanks @HosseinYousefi ! 🚀

PR #371, opened 2023-08-24
Remove APK builds from windows minimal CI

HosseinYousefi:
Somehow building APKs on windows is really slow. macOS minimal tests don't have the apk build, it's not really necessary to build the APKs in windows as the build system relies on Gradle and Flutter not something jnigen specific.

We already have tests for building Windows Flutter apps.


dcharkes (approved):
lgtm

PR #376, opened 2023-08-28
Dart exceptions

### LINKED ISSUE #70: Support implementing Java interfaces with Dart classes
### Implementation strategy 1: native methods

JNI features a way of declaring native methods.

```java
package p.q.r;

class A {
    native double f(int i, String s);
    // ...
}
```

```c
jdouble Java_p_q_r_A_f__ILjava_lang_String_2 (
     JNIEnv *env,        /* interface pointer */
     jobject obj,        /* "this" pointer */
     jint i,             /* argument #1 */
     jstring s)          /* argument #2 */
{
     /* Obtain a C-copy of the Java string */
     const char *str = (*env)->GetStringUTFChars(env, s, 0);

     /* process the string */
     ...

     /* Now we are done with str */
     (*env)->ReleaseStringUTFChars(env, s, str);

     return ...
}
```

https://docs.oracle.com/en/java/javase/17/docs/specs/jni/design.html

We should consider if it makes sense to add some kind of way to generate a Java class that delegates all of its methods to Dart.

This would generate:

* Java class implementing a set of interfaces.
* C file forwarding all methods to Dart callbacks. The Dart callbacks need to be done to a Dart object.
* Dart interface with all the signatures.

The developer would then implement the Dart interface.

### Implementation strategy 2: proxies

https://docs.oracle.com/javase/8/docs/technotes/guides/reflection/proxy.html

Pro:

* Possibly no need to generate (and compile) C code.

Con:

* Might be less performant.

### Related issues

The more general version of:

* https://github.com/dart-lang/jnigen/issues/65

HosseinYousefi:
Closes #368. Final piece of the puzzle to close #70.

When the Dart code implementing the Java interface throws an error. We construct a special `DartException` and return that through the invoke function. We then check if the object is of type `DartException` and if so, we throw it instead of returning.

Q: But what if I have an interface with a method that returns `DartException`, how can I differentiate between returning and throwing? 

A: Don't!

Better A: It's now private and final!

Q: What if we want to also write Java code to handle the Dart exception? 

A: We can revert the last commit.


dcharkes:
> When the Dart code implementing the Java interface throws an error. We construct a special `DartException` and return that through the invoke function. We then check if the object is of type `DartException` and if so, we throw it instead of returning.
> 
> Q: But what if I have an interface with a method that returns `DartException`, how can I differentiate between returning and throwing?
> 
> A: Don't!
> 
> Better A: It's now private and final!
> 
> Q: What if we want to also write Java code to handle the Dart exception?
> 
> A: We can revert the last commit.

If we create a new class, `_ThrownDartException` with a field that contains the object, we can differentiate between us wrapping an exception to return in a handle rather than having an uncaught exception propagate. If we also expose an interface `ThrownDartException` that has a getter for the field, users can unwrap the exception in a try-catch in Java. @HosseinYousefi WDYT about just supporting these use cases right away?

HosseinYousefi:
`ThrownDartException` is called `DartException` right now and it exists. And I can revert back the last two commits to make it public as well.

If this special exception is not thrown by some interface method, then it will throw an `UndeclaredThrowableException` and puts the actual `DartException` nested in the `cause` of this exception.

We could of course change the machinery to have another boolean field `isException` that needs to also be returned every time. But I think this is an overkill, since who will realistically make a custom Java interface with a method that returns `DartException`? 

To fully prevent this I made it private, but I guess a case could be made to have it as public so custom code can at least detect it or mark any custom interface that is known to be implemented in Dart with `throws DartException`.

dcharkes:
In the case where we have existing Java/Kotlin code there wont be any code interacting with our custom exception, because it doesn't know about it.

In the case that someone develops Java/Kotlin code, knowing about the the effective API to call Dart methods from the JVM, they might want to catch the exception and do something with it.

I guess only in the case where someone develops some kind of framework on top of JNIgen would they possibly want to pass those exception objects around. So, maybe my suggestion is overkill.

Is that a good assessment of the situation?

dcharkes:
Alright. We can leave some open issue for the fancy use cases.

jnigen/test/simple_package_test/runtime_test_registrant.dart
```diff
@@ -624,6 +624,47 @@ void registerTests(String groupName, TestRunnerCallback test) {
...
       });
     }
+    group('Dart exceptions are handled', () {
+      for (final sameThread in [true, false]) {
+        test('on ${sameThread ? 'the same thread' : 'another thread'}',
+            () async {
+          final runnable = MyRunnable.implement(
+            $MyRunnableImpl(
+              run: () {
+                throw UnimplementedError();
```
dcharkes, line 634: Maybe also throw a String or something, in Dart anything can be thrown, not just Error and Exception subtypes. It would be good to exercise that.

PR #390, opened 2023-09-11
Add `JBuffer`, `JByteBuffer`

HosseinYousefi:
Speeds up working with buffers. 

@brianquinlan Here's a lean version of `JBuffer` and `JByteBuffer` APIs. Some changes have been made to the Java API to make it more Dart like. Such as making some methods return `void` due to the chaining operator in Dart (`..`) and using getters and setters instead of the traditional names. 

Some other methods have been left out which seemed not as useful. They can always be added in a later version.

TODOs:

- [x] Tests
- [x] Adding symbols to jni_symbols.yaml
- [x] Add a `fromList` factory to `JByteBuffer`

### DIFF
--- a/jni/lib/src/nio/jbyte_buffer.dart
+++ b/jni/lib/src/nio/jbyte_buffer.dart
@@ -47,21 +47,35 @@ final class JByteBufferType extends JObjType<JByteBuffer> {
 /// allocated (See [JByteBuffer.allocateDirect]).
 ///
 /// To create a [JByteBuffer] from the content of a [Uint8List],
-/// use [JByteBuffer.fromList]. This uses direct allocation and will greatly
-/// speed up the copying.
+/// use [JByteBuffer.fromList]. This uses direct allocation to enable fast
+/// copying.
 ///
 /// [asUint8List] provides a direct access to the underlying [Uint8List] that
 /// this buffer uses. This means any changes to it will change the content of
-/// the buffer and vice versa. You can use this to access to [Uint8List] methods
+/// the buffer and vice versa. This can be used to access to [Uint8List] methods
 /// such as [Uint8List.setRange].
 ///
-/// It is safe to use [asUint8List]. As long as the resulting list is
-/// accessible, the associated Java buffer will also be kept alive. Once all the
-/// instances of the original buffer and the lists produced from [asUint8List]
-/// are inaccessible both in Java and Dart, Java will correctly garbage collects
-/// the buffer and frees its underlying memory.
+/// Example:
+/// ```dart
+/// final directBuffer = JByteBuffer.allocateDirect(3);
+/// directBuffer.asUint8List().setAll(0, [1, 2, 3]);
+/// // The buffer is now 1, 2, 3.
+/// ```
 ///
-/// You can choose to [release] the original buffer when calling [asUint8List]
+/// Both the original buffer and the [Uint8List] keep the underlying Java buffer
+/// alive. Once all the instances of the original buffer and the lists produced
+/// from [asUint8List] are inaccessible both in Java and Dart, Java will
+/// correctly garbage collects the buffer and frees its underlying memory.
+///
+/// Example:
+/// ```dart
+/// final directBuffer = JByteBuffer.allocateDirect(3);
+/// final data = directBuffer.asUint8List();
+/// directBuffer.release(); // Releasing the original buffer.
+/// data.setAll(0, [1, 2, 3]); // Works! [data] is still accessible.
+/// ```
+///
+/// The original buffer can be [release]d when calling [asUint8List]
 /// by setting the `releaseOriginal` parameter to `true`.
 ///
 /// Example:
@@ -72,7 +86,7 @@ final class JByteBufferType extends JObjType<JByteBuffer> {
 /// directBuffer.nextByte = 42; // No problem!
 /// print(data1[0]); // prints 42!
 /// final data2 = directBuffer.asUint8List(releaseOriginal: true);
-/// // directBuffer.nextByte = 42; // throws[UseAfterReleaseException]!
+/// // directBuffer.nextByte = 42; // throws [UseAfterReleaseException]!
 /// ```
 class JByteBuffer extends JBuffer {
   @override
--- a/jni/lib/src/nio/jbyte_buffer.dart
+++ b/jni/lib/src/nio/jbyte_buffer.dart
@@ -9,6 +9,7 @@ import '../accessors.dart';
 import '../jarray.dart';
 import '../jni.dart';
 import '../jprimitives.dart';
+import '../jreference.dart';
 import '../jvalues.dart';
 import '../third_party/generated_bindings.dart';
 import '../types.dart';
@@ -60,9 +61,13 @@ class JByteBuffer extends JBuffer {
   /// Throws:
   /// * [IllegalArgumentException] - If the capacity is a negative integer
   factory JByteBuffer.allocateDirect(int capacity) {
-    return const JByteBufferType().fromRef(Jni.accessors
-        .callStaticMethodWithArgs(_class.reference, _allocateDirectId,
-            JniCallType.objectType, [JValueInt(capacity)]).object);
+    return JByteBuffer.fromRef(
+      Jni.accessors.callStaticMethodWithArgs(
+          _class.reference,
+          _allocateDirectId,
+          JniCallType.objectType,
+          [JValueInt(capacity)]).object,
+    );
   }
 
   static final _allocateId = Jni.accessors.getStaticMethodIDOf(
@@ -184,18 +189,39 @@ class JByteBuffer extends JBuffer {
 
   /// Returns this byte buffer as a [Uint8List].
   ///
-  /// Throws [ArgumentError] if the buffer is not direct or the JVM does not
-  /// support the direct buffer operations.
-  Uint8List asUint8List() {
-    assert(isDirect, 'The buffer must be direct.');
+  /// If [releaseOriginal] is `true`, this byte buffer will be released.
+  ///
+  /// Throws [StateError] if the buffer is not direct or the JVM does not
+  /// support the direct buffer operations or the object is an unaligned view
+  /// buffer and the processor does not support unaligned access.
+  Uint8List asUint8List({bool releaseOriginal = false}) {
+    if (!isDirect) {
+      throw StateError(
+        'The buffer must be created with `JByteBuffer.allocateDirect`.',
+      );
+    }
     final address = Jni.env.GetDirectBufferAddress(reference);
     if (address == nullptr) {
-      throw ArgumentError();
+      StateError(
+        'The memory region is undefined or '
+        'direct buffer access is not supported by this JVM.',
+      );
     }
     final capacity = Jni.env.GetDirectBufferCapacity(reference);
     if (capacity == -1) {
-      throw ArgumentError();
+      StateError(
+        'The object is an unaligned view buffer and the processor '
+        'architecture does not support unaligned access.',
+      );
     }
-    return address.cast<Uint8>().asTypedList(capacity);
+    final token = releaseOriginal ? reference : Jni.env.NewGlobalRef(reference);
+    if (releaseOriginal) {
+      setAsReleased();
+    }
+    return address.cast<Uint8>().asTypedList(
+          capacity,
+          token: token,
+          finalizer: Jni.env.ptr.ref.DeleteGlobalRef.cast(),
+        );
   }
 }
--- a/jni/lib/src/nio/jbyte_buffer.dart
+++ b/jni/lib/src/nio/jbyte_buffer.dart
@@ -15,7 +15,7 @@ import '../third_party/generated_bindings.dart';
 import '../types.dart';
 import 'jbuffer.dart';
 
-class JByteBufferType extends JObjType<JByteBuffer> {
+final class JByteBufferType extends JObjType<JByteBuffer> {
   const JByteBufferType();
 
   @override
--- a/jni/lib/src/nio/jbyte_buffer.dart
+++ b/jni/lib/src/nio/jbyte_buffer.dart
@@ -120,6 +120,13 @@ class JByteBuffer extends JBuffer {
     );
   }
 
+  /// Creates a [JByteBuffer] from the content of [list].
+  factory JByteBuffer.fromList(List<int> list) {
+    final buffer = JByteBuffer.allocateDirect(list.length);
+    buffer._asUint8ListUnsafe().setAll(0, list);
+    return buffer;
+  }
+
   static final _sliceId = Jni.accessors
       .getMethodIDOf(_class.reference, r"slice", r"()Ljava/nio/ByteBuffer;");
 
@@ -187,6 +194,43 @@ class JByteBuffer extends JBuffer {
             reference, _arrayId, JniCallType.objectType, []).object);
   }
 
+  void _ensureIsDirect() {
+    if (!isDirect) {
+      throw StateError(
+        'The buffer must be created with `JByteBuffer.allocateDirect`.',
+      );
+    }
+  }
+
+  Pointer<Void> _directBufferAddress() {
+    final address = Jni.env.GetDirectBufferAddress(reference);
+    if (address == nullptr) {
+      StateError(
+        'The memory region is undefined or '
+        'direct buffer access is not supported by this JVM.',
+      );
+    }
+    return address;
+  }
+
+  int _directBufferCapacity() {
+    final capacity = Jni.env.GetDirectBufferCapacity(reference);
+    if (capacity == -1) {
+      StateError(
+        'The object is an unaligned view buffer and the processor '
+        'architecture does not support unaligned access.',
+      );
+    }
+    return capacity;
+  }
+
+  Uint8List _asUint8ListUnsafe() {
+    _ensureIsDirect();
+    final address = _directBufferAddress();
+    final capacity = _directBufferCapacity();
+    return address.cast<Uint8>().asTypedList(capacity);
+  }
+
   /// Returns this byte buffer as a [Uint8List].
   ///
   /// If [releaseOriginal] is `true`, this byte buffer will be released.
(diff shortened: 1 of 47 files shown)


jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,201 @@
...
+        .callMethodWithArgs(
+            reference, _arrayId, JniCallType.objectType, []).object);
+  }
+
+  /// Returns this byte buffer as a [Uint8List].
+  ///
+  /// Throws [ArgumentError] if the buffer is not direct or the JVM does not
+  /// support the direct buffer operations.
+  Uint8List asUint8List() {
+    assert(isDirect, 'The buffer must be direct.');
```
brianquinlan, line 190: if (!isDirect) {
  throw StateError('The buffer must be created with JByteBuffer.allocateDirect');
}

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,201 @@
...
+    final address = Jni.env.GetDirectBufferAddress(reference);
+    if (address == nullptr) {
+      throw ArgumentError();
+    }
+    final capacity = Jni.env.GetDirectBufferCapacity(reference);
+    if (capacity == -1) {
+      throw ArgumentError();
```
brianquinlan, line 197: StateError('The object is an unaligned view buffer and the processor architecture does not support unaligned access');

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,201 @@
...
+      Jni.accessors.getMethodIDOf(_class.reference, r"array", r"()[B");
+
+  @override
+  JArray<jbyte> get array {
+    return const JArrayType(jbyteType()).fromRef(Jni.accessors
+        .callMethodWithArgs(
+            reference, _arrayId, JniCallType.objectType, []).object);
+  }
+
+  /// Returns this byte buffer as a [Uint8List].
```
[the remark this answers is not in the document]
HosseinYousefi: > Eventually we could make it so that JByteBuffer implements List<Uint8>.

Yeah, this is nicer. I'll go with it.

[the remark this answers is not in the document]
dcharkes: DBC

>  if the UInt8List is used after the java.nio.ByteBuffer is collected then you will be accessing deallocated memory

This is the typical reason why we have `Finalizable`s that wrap `Pointer`s.

To avoid extra copying, we don't want to eagerly copy from Java to Dart memory. So indeed, this class should indeed have the methods that use the `address` internally.

And it would make sense to have this class then implement `List<int>` indeed. 👍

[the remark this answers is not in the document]
HosseinYousefi: From our discussion with @dcharkes:

`.asTypedList` has `finalizer` and `token` parameters and whenever the object is GCd, `finalizer` will be invoked for `token`. We can use this mechanism to create a safe `asUint8List` that removes a Java global reference when GCd.

`asUint8List` will also have an `releaseOriginal = false` flag that – when `true` – reuses the same global reference and sets the original object as released.

HosseinYousefi:
I also included a `_asUint8ListUnsafe()` which does **not** attach any finalizer and can be used to add direct useful methods to `JByteBuffer`. We could consider having this:

Implement `List<int>` in `JByteBuffer`
1. With all methods using `_asUint8ListUnsafe()` under the hood and throwing error if the buffer is indirect.
2. With all methods checking `if (isDirect)` and handling the indirect case in a less efficient manner.

Right now I thought it's enough to provide `asUint8List()`, so e.g. `setRange` could be achieved like `buffer.asUint8List().setRange(...)` but this can make it shorter (and a tiny bit faster, if you want to keep the `JBuffer`) into `buffer.setRange(...)`.

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,262 @@
...
+    return address.cast<Uint8>().asTypedList(
+          capacity,
+          token: token,
+          finalizer: Jni.env.ptr.ref.DeleteGlobalRef.cast(),
+        );
+  }
+}
+
+extension Uint8ListToJava on Uint8List {
+  /// Creates a [JByteBuffer] from the content of this list.
```
brianquinlan, line 258: The [JByteBuffer] will be allocated using [JByteBuffer.allocateDirect].

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,262 @@
...
+      Jni.accessors.callStaticMethodWithArgs(
+        _class.reference,
+        _wrapId,
+        JniCallType.objectType,
+        [array.reference, JValueInt(offset), JValueInt(length)],
+      ).object,
+    );
+  }
+
+  /// Creates a [JByteBuffer] from the content of [list].
```
brianquinlan, line 123: The [JByteBuffer] will be created using [JByteBuffer.allocateDirect].

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,262 @@
...
+    final address = _directBufferAddress();
+    final capacity = _directBufferCapacity();
+    return address.cast<Uint8>().asTypedList(capacity);
+  }
+
+  /// Returns this byte buffer as a [Uint8List].
+  ///
+  /// If [releaseOriginal] is `true`, this byte buffer will be released.
+  ///
+  /// Throws [StateError] if the buffer is not direct or the JVM does not
```
brianquinlan, line 238: is not direct (see [JByteBuffer.allocateDirect]) or the ...

jni/pubspec.yaml
```diff
@@ -4,7 +4,7 @@
 
 name: jni
 description: A library to access JNI from Dart and Flutter that acts as a support library for package:jnigen.
-version: 0.6.1
+version: 0.7.0-wip
```
brianquinlan, line 7: Could you do a release so that I can regenerate the files for `cronet_http`? We can also do that in a follow-up PR.

brianquinlan (changes requested):

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,262 @@
...
+      throw StateError(
+        'The buffer must be created with `JByteBuffer.allocateDirect`.',
+      );
+    }
+  }
+
+  Pointer<Void> _directBufferAddress() {
+    final address = Jni.env.GetDirectBufferAddress(reference);
+    if (address == nullptr) {
+      StateError(
```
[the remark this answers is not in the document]
HosseinYousefi: oops!

brianquinlan:
Nice! This is much better than the crap that I wrote originally ;-)

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,267 @@
...
+  @override
+  int get hashCode => (JByteBufferType).hashCode;
+
+  @override
+  bool operator ==(Object other) {
+    return other.runtimeType == (JByteBufferType) && other is JByteBufferType;
+  }
+}
+
+class JByteBuffer extends JBuffer {
```
dcharkes, line 42: This could benefit from a long doc comment:

* What it is (single sentence)
* Why to use it: fast memory copying between Java and Dart with `.setRange`.
* Example use snippets
   * How to use it with `asUint8List` with releaseOriginal true
   * How to use it with `asUint8List` with releaseOriginal false

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,267 @@
...
+        JniCallType.objectType,
+        [array.reference, JValueInt(offset), JValueInt(length)],
+      ).object,
+    );
+  }
+
+  /// Creates a [JByteBuffer] from the content of [list].
+  ///
+  /// The [JByteBuffer] will be allocated using [JByteBuffer.allocateDirect].
+  factory JByteBuffer.fromList(List<int> list) {
```
dcharkes, line 126: I think it's better to use Uint8List as a type here. I don't believe the `setAll` will be fast with `List<int>`.

Yes, it makes it slightly less convenient for users, but it will be fast. And then it's explicit for users that getting things from a generic `List<int>` is slow.

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,267 @@
...
+  /// buffer's content.
+  JByteBuffer slice() {
+    return const JByteBufferType().fromRef(Jni.accessors.callMethodWithArgs(
+        reference, _sliceId, JniCallType.objectType, []).object);
+  }
+
+  static final _duplicateId = Jni.accessors.getMethodIDOf(
+      _class.reference, r"duplicate", r"()Ljava/nio/ByteBuffer;");
+
+  /// Creates a new byte buffer that shares this buffer's content.
```
dcharkes, line 145: This could benefit from some explanation how this behaves as seen from a Dart programmer perspective.

It calls to Java to do `duplicate`.

Can we add the full documentation from Java instead of only the first line?

https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#duplicate--

I'm not entirely sure how to combine the original docs from Java and the line that from Dart's perspective it's implemented by calling the Java method.

```dart
  /// Creates a new byte buffer that shares this buffer's content.
  ///
  /// Calls the Java method `duplicate` on this object.
  ///
  /// The content of the new buffer will be that of this buffer. Changes to this
  /// buffer's content will be visible in the new buffer; the new buffer itself,
  /// however, will be read-only and will not allow the shared content to be
  /// modified. The two buffers' position, limit, and mark values will be
  /// independent.
  ///
  /// The new buffer's capacity, limit, position, and mark values will be
  /// identical to those of this buffer.
  ///
  /// If this buffer is itself read-only then this method behaves in exactly the
  /// same way as the
  /// [duplicate](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#duplicate--)
  /// method.
  JByteBuffer duplicate() {
    return const JByteBufferType().fromRef(Jni.accessors.callMethodWithArgs(
        reference, _duplicateId, JniCallType.objectType, []).object);
  }
```

or

```dart
  /// Creates a new byte buffer that shares this buffer's content. (Calls the
  /// Java method `duplicate` on this object.)
  ///
  /// The content of the new buffer will be that of this buffer. Changes to this
  /// buffer's content will be visible in the new buffer; the new buffer itself,
  /// however, will be read-only and will not allow the shared content to be
  /// modified. The two buffers' position, limit, and mark values will be
  /// independent.
  ///
  /// The new buffer's capacity, limit, position, and mark values will be
  /// identical to those of this buffer.
  ///
  /// If this buffer is itself read-only then this method behaves in exactly the
  /// same way as the
  /// [duplicate](https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html#duplicate--)
  /// method.
  JByteBuffer duplicate() {
    return const JByteBufferType().fromRef(Jni.accessors.callMethodWithArgs(
        reference, _duplicateId, JniCallType.objectType, []).object);
  }
```

ditto for other methods from Java.

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,262 @@
...
+  /// Returns this byte buffer as a [Uint8List].
+  ///
+  /// If [releaseOriginal] is `true`, this byte buffer will be released.
+  ///
+  /// Throws [StateError] if the buffer is not direct or the JVM does not
+  /// support the direct buffer operations or the object is an unaligned view
+  /// buffer and the processor does not support unaligned access.
+  Uint8List asUint8List({bool releaseOriginal = false}) {
+    _ensureIsDirect();
+    final address = _directBufferAddress();
```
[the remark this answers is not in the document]
dcharkes: lgtm!

jni/test/jbyte_buffer_test.dart
```diff
@@ -0,0 +1,213 @@
...
+      expect(buffer.remaining, 0);
+      expect(buffer.hasRemaining, false);
+    });
+  });
+
+  testRunner('isReadOnly and asReadOnlyBuffer', () {
+    using((arena) {
+      final buffer = testDataBuffer(arena);
+      expect(buffer.isReadOnly, false);
+      final readOnly = buffer.asReadOnlyBuffer();
```
dcharkes, line 140: `final readOnly = buffer.asReadOnlyBuffer()..releasedBy(arena);` ?

jni/test/jbyte_buffer_test.dart
```diff
@@ -0,0 +1,213 @@
...
+      expect(sliced.nextByte, 2);
+    });
+  });
+
+  testRunner('duplicate', () {
+    using((arena) {
+      final buffer = testDataBuffer(arena);
+      buffer.position = 1;
+      buffer.limit = 2;
+      final duplicate = buffer.duplicate();
```
dcharkes, line 181: `..releasedBy(arena)`?

jnigen/lib/src/tools/android_sdk_tools.dart
```diff
@@ -202,6 +202,7 @@ task $_gradleGetSourcesTaskName(type: Copy) {
       File(buildGradleOld).deleteSync();
     }
     if (procRes.exitCode != 0) {
+      print(procRes.stdout);
```
dcharkes, line 205: left over debug print?

jnigen/lib/src/bindings/dart_generator.dart
```diff
@@ -589,7 +589,7 @@ class _$implClassName$typeParamsDef implements $implClassName$typeParamsCall {
         ? '($typeClassName).hashCode'
         : 'Object.hash($typeClassName, $hashCodeTypeClasses)';
     s.write('''
-class $typeClassName$typeParamsDef extends $_jType<$name$typeParamsCall> {
+final class $typeClassName$typeParamsDef extends $_jType<$name$typeParamsCall> {
```
dcharkes, line 592: Nit: could have been a separate PR to ease reviewing. (I have already reviewed now, so keep it in here!)

jnigen/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## 0.6.1-wip
-* Add `ignore_for_file: lines_longer_than_80_chars` to the generated file preamble.
+## 0.7.0
+* **Breaking Change** ([#387](https://github.com/dart-lang/jnigen/issues/387)): Added `JBuffer` and `JByteBuffer` classes as default classes for `java.nio.Buffer` and `java.nio.ByteBuffer` respectively.
```
dcharkes, line 2: nit: line wrap at 80?

jni/lib/src/util/jlist.dart
```diff
@@ -196,7 +196,7 @@ class JList<$E extends JObject> extends JObject with ListMixin<$E> {
           reference,
           _insertAllId,
           JniCallType.booleanType,
-          [JValueInt(index), (iterable as JObject).reference]);
+          [JValueInt(index), (iterable as JObject).reference]).boolean;
```
dcharkes, line 199: nit: prefer trailing comma for multi-line arguments.

dcharkes (approved):
lgtm!

Please add also a speedup factor to the PR description, our motivation! 💨

jni/lib/src/nio/jbyte_buffer.dart
HosseinYousefi: I'll address this in another PR. Other classes like also don't have any documentation. 

https://github.com/dart-lang/jnigen/issues/391

jni/lib/src/nio/jbyte_buffer.dart
HosseinYousefi: Fair enough.

jni/lib/src/nio/jbyte_buffer.dart
HosseinYousefi: Btw the docs you have here is for `asReadOnlyBuffer`.

HosseinYousefi:
@dcharkes Please review the added docs.

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+
+/// A byte [JBuffer].
+///
+/// The bindings for `java.nio.ByteBuffer`.
+///
+/// This enables fast memory copying between Java and Dart when directly
+/// allocated (See [JByteBuffer.allocateDirect]).
+///
+/// To create a [JByteBuffer] from the content of a [Uint8List],
+/// use [JByteBuffer.fromList]. This uses direct allocation and will greatly
```
dcharkes, line 50: Maybe rephrase as "This uses direct allocation to enable fast copying".

Speed up is a comparison, but a comparison to what?

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+/// speed up the copying.
+///
+/// [asUint8List] provides a direct access to the underlying [Uint8List] that
+/// this buffer uses. This means any changes to it will change the content of
+/// the buffer and vice versa. You can use this to access to [Uint8List] methods
```
dcharkes, line 55: Avoid "you". This can be used to ...

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+/// such as [Uint8List.setRange].
+///
+/// It is safe to use [asUint8List]. As long as the resulting list is
```
dcharkes, line 58: The first sentence saying "safe" makes me wonder safe in what sense before I read on. Maybe rephrase to: "Both the original buffer and the Uint8List keep the underlying Java buffer alive. Once all the ..." (I think the right terms to use are "alive" and "release". Maybe don't use "safe" but spell it out that we will not try to read the Java memory from Dart after it has been "released".)

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+/// accessible, the associated Java buffer will also be kept alive. Once all the
+/// instances of the original buffer and the lists produced from [asUint8List]
+/// are inaccessible both in Java and Dart, Java will correctly garbage collects
+/// the buffer and frees its underlying memory.
+///
+/// You can choose to [release] the original buffer when calling [asUint8List]
```
dcharkes, line 64: another you

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+///
+/// To create a [JByteBuffer] from the content of a [Uint8List],
+/// use [JByteBuffer.fromList]. This uses direct allocation and will greatly
+/// speed up the copying.
+///
+/// [asUint8List] provides a direct access to the underlying [Uint8List] that
+/// this buffer uses. This means any changes to it will change the content of
+/// the buffer and vice versa. You can use this to access to [Uint8List] methods
+/// such as [Uint8List.setRange].
+///
```
dcharkes, line 57: I'd move the simplest version of the example here.

Please make the example also memory safe with using the arena or releasing manually. Users are going to copy and paste it.

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+/// It is safe to use [asUint8List]. As long as the resulting list is
+/// accessible, the associated Java buffer will also be kept alive. Once all the
+/// instances of the original buffer and the lists produced from [asUint8List]
+/// are inaccessible both in Java and Dart, Java will correctly garbage collects
+/// the buffer and frees its underlying memory.
+///
```
dcharkes, line 63: Then a second example here, where we release the buffer early, and access the Uint8List after.

jni/lib/src/nio/jbyte_buffer.dart
```diff
@@ -0,0 +1,302 @@
...
+/// You can choose to [release] the original buffer when calling [asUint8List]
+/// by setting the `releaseOriginal` parameter to `true`.
+///
```
dcharkes, line 66: Then third example here where we use `releaseOriginal: true`.

dcharkes (approved):

PR #399, opened 2023-09-20
WIP: Use `super.reference` in `fromRef` constructors

HosseinYousefi:
Minor refactoring to use the newer Dart features in both the generated code and in `package:jni` classes.

WIP: There is still some issues with subclasses in the generated code.


HosseinYousefi:
Closing this as using `super.reference` would required `fromRef` to accept required named parameters for the type classes (similar to other constructors and methods). However this is a larger change than what I originally intended and it will not benefit the user since `fromRef` is rarely used directly.

Issue #31, opened 2022-08-26
String escaping


mahesh-hegde:
Currently we are pasting integer, float and String constants in java classes into dart classes directly. There are few edge cases though, where some non-escapable characters get included.

1) Escape string on java side: I should pull apache commons text for this one purpose, String escape doesn't seem to be in standard library.

2) Escape in dart: same story here. There's a strings package but that seems to be discontinued.

What I would ideally have is something like https://pkg.go.dev/strconv#Quote

Otherwise should code this logic manually.

mahesh-hegde:
We can think this issue from other direction.

Currently, static fields are not binded through FFI, instead they are translated as dart const static fields. That's why this issue arises.

But with strings, we probably want a getter, because it's rare to use java string literal value in dart (by my experience). But when we pass such static string field to Java code, we need to convert it to Java string.

So maybe it's better to leave it as a getter and not convert to direct literal.

cc: @dcharkes .

dcharkes:
Good observation. Let's see what is most common in practise. If both are common, we can also generate both. (The Dart consts will be tree-shaken anyway in AOT.)

dcharkes:
One benefit of having them in Dart is easier reasoning about your code when writing code in Dart. E.g. doing a switch or equality check on something. When in the debugger you can see the string contents.

HosseinYousefi:
We can generate the strings as raw strings like `r"content"`. So we won't need to have any escaping logic.

HosseinYousefi:
> I am afraid it may not cover all cases. For example, java string contains escaped character like \n or \r.

`r"\n"` is basically `"\\n"`, right? What would be the issue here?

mahesh-hegde:
Here's an example.

https://github.com/dart-lang/jnigen/blob/e492c8f486171aebc67def75d54f3a02390c94de/jnigen/example/pdfbox_plugin/jnigen.yaml#L58

If you google the javadoc for excluded field, you get the string literal `\"`, which java parser turns into `"`. It passes through JSON library which has handling for all these cases, we get `"` in dart, we try to wrap it in quotes. But that's an invalid dart string literal.

If you use single quotes to wrap strings, you will get same issue for a string `\'`.

So the plausible solutions are:

1. Implement escaping ourselves

2. expose static final string fields like any other java field, i.e through runtime field getter.

with trade offs discussed above.

HosseinYousefi:
[...]

PR #400, opened 2023-09-25
Add `JLazyReference` and `JFinalString`

HosseinYousefi:
This is going to be used for `static final String` fields. These classes are never directly exposed to the end users and are only visible for jnigen-generated code.

As discussed [here](https://github.com/dart-lang/jnigen/issues/31#issuecomment-1733660665); `JFinalString` has a defined value and `toDartString` uses that value. In case the user simply wants the string value and does not use the Java reference, the method obtaining the reference is never going to be used. This works because of `JLazyReference`.

Once this reviewed and merged, I will open a second PR, to actually generate the `JString get xyz => JFinalString(...)` for `static final String` fields in `package:jnigen`.

### DIFF
--- a/jni/lib/src/jfinal_string.dart
+++ b/jni/lib/src/jfinal_string.dart
@@ -1,0 +1,31 @@
+// Copyright (c) 2023, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:ffi';
+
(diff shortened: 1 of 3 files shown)


dcharkes (approved):
lgtm

I was expecting changes in package:jnigen as well to generate `JFinalString`s.

HosseinYousefi:
> lgtm
> 
> I was expecting changes in package:jnigen as well to generate `JFinalString`s.

I'll do that in another PR.

PR #418, opened 2023-11-16
Fix beta CI and print the analyzer failures

HosseinYousefi:
Two issues that are solved:

1. There is a new lint called `doc_directive_unknown` enabled by default which does not recognize elements like `@code` in the documentation. Now generating an ignore for it.
2. Not generating the latest android version code.


dcharkes (approved):
LGTM!

